PR #8, opened 2020-04-24
protocolManager support for React Native

radko93:
I didn't want to change too much so I made sure default crypto is detected correctly. This has to be tested on the web.


lib/utils.js
```diff
@@ -3,7 +3,11 @@ export function getGlobalScope() {
 }
 
 export function isWebEnvironment()  {
-  return getGlobalScope() !== null;
+  return typeof document != 'undefined';
```
mobitar, line 6: Should be `!==` or intentionally using `!=`?

radko93: Typo, I update that now.

PR #10, opened 2020-05-05
Webpack configuration

johnny243:


webpack.dev.js
```diff
@@ -0,0 +1,15 @@
+const merge = require('webpack-merge');
+const config = require('./webpack.config.js');
+const WriteFilePlugin = require('write-file-webpack-plugin');
+
+module.exports = merge(config, {
+  mode: 'development',
+  devtool: 'source-map',
+  plugins: [
+    ...config.plugins,
+    new WriteFilePlugin()
```
baptiste-grob, line 10: You shouldn't need a plugin to write your bundle to disk when running `webpack -w` without the dev server. In this case it will have no effect

webpack.dev.js
```diff
@@ -0,0 +1,15 @@
...
+  ],
+  stats: {
+    colors: true
```
baptiste-grob, line 13: 🌈

webpack.dev.js
```diff
@@ -0,0 +1,15 @@
+const merge = require('webpack-merge');
+const config = require('./webpack.config.js');
+const WriteFilePlugin = require('write-file-webpack-plugin');
+
+module.exports = merge(config, {
+  mode: 'development',
+  devtool: 'source-map',
+  plugins: [
+    ...config.plugins,
```
baptiste-grob, line 9: Maybe I'm wrong but I believe you don't need to spread anything when using webpack-merge

johnny243: You're correct :wink:

webpack.dev.js
johnny243: Yeah, I kept it from the original `webpack.js` file :smile:

webpack.dev.js
johnny243: We are using `webpack-dev-server` instead of `webpack` for the start script. See: https://github.com/standardnotes/web/pull/398#discussion_r420507179

baptiste-grob: I understood that for the web repo but for snjs it seems we're only using vanilla webpack: https://github.com/standardnotes/snjs/blob/354dee8a881318492bf466370a6f200564577c89/package.json#L7

PR #11, opened 2020-05-08
Session management integration

johnny243:
I improved from my previous solution (which seemed a bit _hacky_) and instead used existing features from `snjs`:

- A custom HTTP status code is returned from the server when an expired token is provided to the request. This will prevent the app from triggering the `InvalidSyncSession` event, which opens a pop-up to the user.
- ~~The `SyncEvent` triggered by the above condition will execute the refresh session action.~~


lib/services/api/session_manager.ts
```diff
@@ -276,15 +280,51 @@ export class SNSessionManager extends PureService {
...
+  private async handleRefreshSessionResponse(response: HttpResponse) {
+    if (response.error) {
+      return;
+    }
+    const session = this.initializeSessionFromResponse(response);
+    await this.storageService!.setValue(StorageKey.Session, session);
+    await this.setSession(session);
+  }
+
+  private initializeSessionFromResponse(response: HttpResponse) {
```
mobitar, line 323: Should be named `newSessionFromResponse` instead. `initialize` sounds like you're setting state on an existing object, or, you're creating a new session and also holding onto it. But here you just create a detached object and return it.

lib/services/api/session_manager.ts
```diff
@@ -276,15 +280,51 @@ export class SNSessionManager extends PureService {
...
     const user = response.user;
     this.user = user;
     await this.storageService!.setValue(StorageKey.User, user);
-    const session = new Session(response.token);
+    const session = this.initializeSessionFromResponse(response);
     await this.storageService!.setValue(StorageKey.Session, session);
     await this.setSession(session);
   }
+
+  private async handleRefreshSessionResponse(response: HttpResponse) {
```
mobitar, line 314: This function should probably just be inline inside of refreshSession, if it's not reusable.

lib/application.ts
```diff
@@ -1383,4 +1389,21 @@ export class SNApplication {
       return base;
     }
   }
+
+  public async handleExpiredAccessToken() {
```
mobitar, line 1393: Why is the `application` handling expired tokens? This sounds like a service's internal implementation detail. Application functions, and specifically `public` ones, are meant for client consumption (web/mobile).

lib/services/application_service.ts
```diff
@@ -43,6 +43,8 @@ export class ApplicationService extends PureService {
         this.onAppIncrementalSync();
       } else if (event === ApplicationEvent.KeyStatusChanged) {
         this.onAppKeyChange();
+      } else if (event === ApplicationEvent.ExpiredAccessToken) {
+        this.onAppExpiredToken();
```
mobitar, line 47: Not sure I'm following why this is in application_service? This is an abstract class that clients can subclass from. But it's a client related wrapper, not snjs related wrapper. And there can exist many subclasses of this class, so in this case, `onAppExpiredToken` can be called ten times.

lib/services/sync/sync_service.ts
```diff
@@ -794,6 +794,9 @@ export class SNSyncService extends PureService {
     if (response.status === INVALID_SESSION_RESPONSE_STATUS) {
       this.notifyEvent(SyncEvent.InvalidSession);
     }
+    if (response.status === EXPIRED_ACCESS_TOKEN_RESPONSE_STATUS) {
+      this.notifyEvent(SyncEvent.ExpiredAccessToken);
```
mobitar, line 798: It seems like renewing session token should be handled by either sync_service or session_manager directly.

lib/services/api/session.ts
```diff
@@ -1,12 +1,35 @@
...
+  private getExpireAt(): number {
+    return this.expireAt || 0;
+  }
+
+  public canExpire() {
+    return this.getExpireAt() > 0;
+  }
+
+  public isExpired() {
+    if (!this.canExpire())
```
mobitar, line 30: Always use { } brackets in if statements.

lib/services/api/api_service.ts
```diff
@@ -172,6 +174,15 @@ export class SNApiService extends PureService {
     return response;
   }
 
+  async signOut() {
+    const url = await this.path(REQUEST_PATH_LOGOUT);
+    await this.httpService!.postAbsolute(
+      url,
+      {},
```
mobitar, line 181: This should be `undefined`. If postAbsolute doesn't accept undefined, it should be modified to do so.

lib/services/api/api_service.ts
```diff
@@ -234,4 +245,23 @@ export class SNApiService extends PureService {
 
     return response;
   }
+
+  async refreshSession(userUuid?: string) {
```
mobitar, line 249: Is userUuid really optional here? Can `refreshSession` function without this? If it can't, it should throw an exception, or not even be called in the first place.

lib/application.ts
```diff
@@ -1383,4 +1389,21 @@ export class SNApplication {
...
+    }
+    this.refreshingSession = true;
+    this.syncService!.lockSyncing();
+    const { response } = await this.sessionManager!.refreshSession();
+    if (!response.error) {
+      this.syncService!.unlockSyncing();
+      this.syncService!.resetSyncState();
+      await this.syncService!.sync();
+    } else {
+      this.notifyEvent(ApplicationEvent.InvalidSyncSession);
```
mobitar, line 1405: Should you unlockSyncing here?

lib/services/application_service.ts
johnny243: Where should I check for when the `SyncEvent.ExpiredAccessToken` event is triggered? Can I do so from within the `api_service` or `session_manager`?

mobitar: I feel like session renewal should be handed completely by the api_service. If a session is expired, it freezes all sync requests within the api_service (or adds them to a resolve queue). Once the session renews, any pending sync requests are executed serially or concurrently, and their original promise resolves. I don't think we need to get the `application` or `sync_service` involved in session management. Take a look at how we preserve the original promises of functions here: https://github.com/standardnotes/snjs/blob/004/lib/services/model_manager.ts#L127

lib/services/sync/sync_service.ts
johnny243: You're right. I'm checking for the status code on `sync_service`, but it would be more appropriate to do so from `session_manager` because this event could be triggered from other requests other than the sync ones (example `changePassword`).

lib/services/application_service.ts
johnny243: The thing is that both `sync_service` and `session_manager` are not available from `api_service`. I need `sync_service` for toggling lock syncing, and I also need `session_manager` to update the session object with the new tokens.

mobitar: The api_service can definitely lock syncing as well if it wanted to.

mobitar: In this case it seems like the best way is to have api_service and session_manager collaborate, but no one else.

johnny243: > The api_service can definitely lock syncing as well if it wanted to.

Should I lock syncing by checking if the session is being renewed? If so, any syncing requests should be queued and processed later.

johnny243: Does it make sense to use `SyncEvent` within `api_service`? I'm thinking of using `this.notifyEvent(SyncEvent.ExpiredAccessToken)` from `http_service` (I will call this method whenever I get the custom HTTP code that is returned when the access token is expired) and in `api_service`, I could have an observer that will trigger my `refreshSession` functionality whenever the `SyncEvent.ExpiredAccessToken` is notified.

mobitar: Nope, http_service is totally agnostic to the functionings of the app. Should definitely not trigger any events. It is stateless. Only sync_service should trigger sync events.

johnny243: Thanks for pointing me into the right direction. I've pushed new changes :+1:

lib/services/api/api_service.ts
```diff
@@ -234,4 +256,56 @@ export class SNApiService extends PureService {
 
     return response;
   }
+
+  private async checkForExpiredAccessToken(errorResponse: HttpResponse) {
```
mobitar, line 260: `checkForExpiredAccessToken` doesn't imply there are side-effects (refresh session). The two actions taken here should be separated.

lib/services/api/api_service.ts
```diff
@@ -234,4 +256,56 @@ export class SNApiService extends PureService {
...
+    }
+    await this.refreshSession();
+  }
+
+  private async refreshSession() {
+    if (this.refreshingSession) {
+      return;
+    }
+    const user = await this.storageService!.getValue(StorageKey.User);
+    if (!user) {
```
mobitar, line 272: Why would refreshSession be called if user is undefined? It seems like we would never get to this point. This sounds like it should throw an exception instead.

```
if(!user || !user.uuid) {
   throw...
}
```

But again, will refreshSession ever be called in an offline context? If not, this check doesn't need to exist.

lib/services/api/api_service.ts
```diff
@@ -234,4 +256,56 @@ export class SNApiService extends PureService {
...
+      return;
+    }
+    if (!user.uuid) {
+      return;
+    }
+    const userUuid = user.uuid;
+    this.refreshingSession = true;
+    const url = await this.path(REQUEST_PATH_SESSION_REFRESH);
+    const params = this.params({
+      user_uuid: userUuid,
```
mobitar, line 282: Why does the server need a user_uuid? Can't it determine the user based on the access_token or refresh_token?

lib/services/api/session_manager.ts
```diff
@@ -283,8 +286,16 @@ export class SNSessionManager extends PureService {
     const user = response.user;
     this.user = user;
     await this.storageService!.setValue(StorageKey.User, user);
-    const session = new Session(response.token);
+    const session = this.newSessionFromResponse(response);
     await this.storageService!.setValue(StorageKey.Session, session);
     await this.setSession(session);
   }
+
+  private newSessionFromResponse(response: HttpResponse) {
```
mobitar, line 294: This is a duplication of code from api_service. This function looks like it should be a static method in the Session class.

lib/services/api/api_service.ts
```diff
@@ -234,4 +256,56 @@ export class SNApiService extends PureService {
...
+      access_token: this.session!.accessToken,
+      refresh_token: this.session!.refreshToken
+    });
+    const result = await this.httpService!.postAbsolute(
+      url,
+      params
+    ).then(async (response) => {
+      const session = this.newSessionFromResponse(response);
+      await this.storageService!.setValue(StorageKey.Session, session);
```
mobitar, line 291: Persisting a session should occur from either the api_service or the session_manager, not both. We can't be persisting the same object from two different locations. You can have session_manager rely on api_service instead to persist if you want.

lib/application.ts
```diff
@@ -1383,4 +1383,5 @@ export class SNApplication {
       return base;
     }
   }
+
```
mobitar, line 1386: Should be removed

lib/services/sync/sync_service.ts
```diff
@@ -36,7 +36,6 @@ const DEFAULT_MAX_DISCORDANCE = 5;
 const DEFAULT_MAJOR_CHANGE_THRESHOLD = 15;
 const INVALID_SESSION_RESPONSE_STATUS = 401;
 
-
```
mobitar, line 39: Should be removed

lib/services/api/session.ts
```diff
@@ -1,12 +1,35 @@
...
+    this.refreshToken = refreshToken;
   }
-}
\ No newline at end of file
+
+  private getCurrentTime() {
+    return Date.now();
+  }
+
+  private getExpireAt(): number {
```
mobitar, line 21: `: number` is redundant. TypeScript can infer this.

lib/services/api/session.ts
```diff
@@ -1,12 +1,35 @@
...
-    this.token = token;
+  constructor(accessToken: string, expireAt?: number, refreshToken?: string) {
+    this.accessToken = accessToken;
+    this.expireAt = expireAt;
+    this.refreshToken = refreshToken;
   }
-}
\ No newline at end of file
+
+  private getCurrentTime() {
```
mobitar, line 17: This function doesn't appear to be re-used. If it's just use once by `isExpired`, should just be inlined.

lib/services/api/api_service.ts
```diff
@@ -16,9 +16,13 @@ const REQUEST_PATH_REGISTER = '/auth';
 const REQUEST_PATH_LOGIN = '/auth/sign_in';
 const REQUEST_PATH_CHANGE_PW = '/auth/change_pw';
 const REQUEST_PATH_SYNC = '/items/sync';
+const REQUEST_PATH_LOGOUT = '/auth/sign_out';
```
mobitar, line 19: Isn't auth/sign_out just destroying the session? In that case, shouldn't this be a DELETE request to /session?

lib/services/api/messages.ts
```diff
@@ -11,6 +11,11 @@ export const API_MESSAGE_CHANGE_PW_IN_PROGRESS     = 'An existing change passwor
 
 export const API_MESSAGE_FALLBACK_LOGIN_FAIL       = 'Invalid email or password.';
 
+export const API_MESSAGE_GENERIC_TOKEN_REFRESH_FAIL= `A server error occurred while trying to refresh your session.
+                                                      Please try again.`;
+
+export const API_MESSAGE_TOKEN_REFRESH_IN_PROGRESS = 'The session is being updated with new tokens, please standby.';
```
mobitar, line 17: "Your account session is being renewed with the server. Please try your request again."

lib/services/api/api_service.ts
johnny243: The `DELETE` request to `/session` can destroy any session for the current user, except the current session.

lib/services/api/session_manager.ts
```diff
@@ -80,7 +80,7 @@ export class SNSessionManager extends PureService {
 
   private async setSession(session: Session) {
     this.session = session;
-    this.apiService!.setSession(this.session);
+    await this.apiService!.setSession(this.session);
```
mobitar, line 83: We call `setSession` from `initializeFromDisk`, and `apiService.setSession` writes to disk. So we don't want to write to disk when we are loading from disk.

lib/services/api/api_service.ts
```diff
@@ -172,6 +178,15 @@ export class SNApiService extends PureService {
     return response;
   }
 
+  async signOut() {
```
mobitar, line 181: What happens if you attempt to signOut with an expired token?

lib/services/api/api_service.ts
```diff
@@ -16,9 +16,13 @@ const REQUEST_PATH_REGISTER = '/auth';
 const REQUEST_PATH_LOGIN = '/auth/sign_in';
 const REQUEST_PATH_CHANGE_PW = '/auth/change_pw';
 const REQUEST_PATH_SYNC = '/items/sync';
+const REQUEST_PATH_LOGOUT = '/auth/sign_out';
+const REQUEST_PATH_SESSION_REFRESH = '/session/refresh';
 
 const API_VERSION = '20200115';
 
+const EXPIRED_ACCESS_TOKEN_RESPONSE_STATUS = 498;
```
mobitar, line 24: I think this should be moved into http_service. 498 is specific to the HTTP protocol and explicitly meant for expired tokens, so it makes sense that http_service would handle this. Then you can expose a function from http_service called `isErrorResponseExpiredToken` which would replace this class' `expiredAccessToken`.

lib/services/api/api_service.ts
johnny243: This is an authenticated action, so the server will return a `498` status code but `snjs` will not perform the request to refresh the token.

Perhaps we should terminate the session from the server, even if the token is expire.

mobitar: I imagine this case will not be uncommon. A user returning to a device they haven't accessed in months will attempt to sign out to refresh stale data. So we do want a solution to this.

test/session.test.js
```diff
@@ -0,0 +1,84 @@
...
+      email: this.email,
+      password: this.password
+    });
+    await promise;
+
+    // Saving the current session information for later...
+    const sessionBeforeSync = this.application.apiService.session;
+
+    // Waiting enough time for the access token to expire, before performing a new sync request.
+    await Factory.sleep(3);
```
mobitar, line 70: Should not be magic number. Should be based on specific expirey time.

test/session.test.js
```diff
@@ -0,0 +1,84 @@
...
+
+    // Performing a sync request with an expired access token.
+    const sync = this.application.sync(syncOptions);
+    await sync;
```
mobitar, line 74: Should just be `await this.application.sync(syncOptions);` without having to make into separate promise var.

johnny243: Right, I forgot to remove this :+1:

test/session.test.js
```diff
@@ -0,0 +1,83 @@
...
+      email: this.email,
+      password: this.password
+    });
+    await promise;
+
+    // Saving the current session information for later...
+    const sessionBeforeSync = this.application.apiService.session;
+
+    // Waiting enough time for the access token to expire, before performing a new sync request.
+    await Factory.sleep((Math.floor(sessionBeforeSync.expireAt - Date.now() / 1000) + 2));
```
mobitar, line 70: What's the + 2? Why not + 1? + 0.5? This should be made clear.

test/session.test.js
```diff
@@ -0,0 +1,83 @@
...
+
+  it.only('should be refreshed if access token is expired', async function () {
+    this.application = await Factory.signOutApplicationAndReturnNew(this.application);
+
+    const promise = Factory.loginToApplication({
+      application: this.application,
+      email: this.email,
+      password: this.password
+    });
+    await promise;
```
mobitar, line 64: Doesn't need to be isolated.

test/session.test.js
```diff
@@ -0,0 +1,83 @@
...
+    this.password = Uuid.GenerateUuidSynchronously();
+    await Factory.registerUserToApplication({
+      application: this.application,
+      email: this.email,
+      password: this.password
+    });
+    this.signOut = async () => {
+      this.application = await Factory.signOutApplicationAndReturnNew(this.application);
+    };
+    this.signIn = async () => {
```
mobitar, line 37: Are these functions used?

test/session.test.js
```diff
@@ -0,0 +1,83 @@
...
+    expect(this.application.syncService.isOutOfSync()).to.equal(false);
+    const items = this.application.itemManager.items;
+    expect(items.length).to.equal(this.expectedItemCount);
+    const rawPayloads = await this.application.storageService.getAllRawPayloads();
+    expect(rawPayloads.length).to.equal(this.expectedItemCount);
+    await this.application.deinit();
+  });
+
+  it.only('should be refreshed if access token is expired', async function () {
+    this.application = await Factory.signOutApplicationAndReturnNew(this.application);
```
mobitar, line 57: Why sign out if signing back in immediately after? Is this intentional?

test/session.test.js
johnny243: The `+ 2` is just an additional 2 seconds to make sure that enough time has elapsed for the token to expire. Also, I used `Math.floor` function (some milliseconds may be subtracted from the remaining time).

mobitar: These sorts of numbers are dangerous and make test maintenance difficult over the long run. The token expirey should be _exact_, so we should not need to pad. If we do pad, it should be no more than 1ms.

lib/services/api/api_service.ts
johnny243: > A user returning to a device they haven't accessed in months will attempt to sign out to refresh stale data.

I'm not following, why do they need to sign out in this case? If they haven't used the app for several month, only the `access_token` will be expired, but not the `refresh_token`. So if they open the app, the `refresh_token` will be used to extend the session for an additional year.

`access_token` will be valid for 60 days, but the `refresh_token` can be used for up to a year.

mobitar: Not saying that the user is rational ;) In many cases if something goes wrong in general, a user will attempt to sign out and sign back in. This is actually advice our support team will give to users in last-resort cases.

johnny243: Oh, I got you now :smile: 

> Not saying that the user is rational ;)

:laughing:

lib/services/api/api_service.ts
```diff
@@ -58,8 +61,11 @@ export class SNApiService extends PureService {
     return this.host;
   }
 
-  public setSession(session: Session) {
+  public async setSession(session: Session, fromDisk: boolean = false) {
```
mobitar, line 64: It would be more clear if `fromDisk` was renamed to `persist`. Because it's not clear how fromDisk will be used from a caller's perspective. `persist` makes this much more clear. Then if(persist) {save to disk}.

test/session.test.js
```diff
@@ -0,0 +1,160 @@
...
+  afterEach(async function () {
+    expect(this.application.syncService.isOutOfSync()).to.equal(false);
+    const items = this.application.itemManager.items;
+    expect(items.length).to.equal(this.expectedItemCount);
+    const rawPayloads = await this.application.storageService.getAllRawPayloads();
+    expect(rawPayloads.length).to.equal(this.expectedItemCount);
+    await this.application.deinit();
+  });
+
+  function getDelayBeforeNextRequest(currentSession) {
```
mobitar, line 45: I think it would be more clear if you created a resuable async function called `sleepUntilSessionExpires`. It's really not clear by its title what this function does. The new function would calculate session expirey time, and also do the sleeping. This way callers don't have to separate the two.

test/session.test.js
```diff
@@ -0,0 +1,160 @@
...
+    const tokenExpireAt = currentSession.expireAt;
+    const timeRemaining = Math.ceil(tokenExpireAt - Date.now() / 1000);
+    /* 
+      If the token has not expired yet, we will return the remaining time, plus one second.
+      Else, there's no need to add a delay.
+    */
+    return timeRemaining > 0 ? timeRemaining + 1 : 0;
```
mobitar, line 52: I still don't think you should be adding 1s. It seems like you're adding 1s "to be safe", but we don't want to be safe, we want to be precise. What's the difference between 1ms and 1s? A session should be 100% expired 1ms after its expirey timestamp.

test/session.test.js
```diff
@@ -0,0 +1,160 @@
...
+    return application.storageService.getValue(StorageKey.Session);
+  }
+
+  it('should fail when a sync request is perfomed with an expired access token', async function () {
+    const currentSession = this.application.apiService.session;
+
+    const delayBeforeNextRequest = getDelayBeforeNextRequest(currentSession);
+    await Factory.sleep(delayBeforeNextRequest);
+
+    const response = await this.application.apiService.sync([]);
```
mobitar, line 65: What's the empty array for? `sync` doesn't take an array param.

test/session.test.js
```diff
@@ -0,0 +1,160 @@
...
+    expect(response.error.tag).to.equal('expired-access-token');
+
+    expect(response.error).to.have.property('message');
+    expect(response.error.message).to.equal('The provided access token has expired.');
+  }).timeout(10000);
+
+  it('should return the new session in the response when refreshed', async function () {
+    const response = await this.application.apiService.refreshSession();
+
+    expect(response).to.have.property('status');
```
mobitar, line 82: So the standard for how tests should be written should be modeled around newer tests that I've written. Older tests I've written tend to violate the principle of "test only one thing" per test. Here's a newer test for example: https://github.com/standardnotes/snjs/blob/004/test/item_manager.test.js#L120

So in this case you want to create a separate test for ensuring `status` exists. Then, you don't need to check this again for every test.

In this case you have two tests that both check for the same properties, when one test should do.

test/session.test.js
```diff
@@ -0,0 +1,160 @@
...
+    expect(response).to.have.property('session');
+
+    expect(response.session).to.have.property('expire_at');
+    expect(response.session.expire_at).to.be.a('number');
+
+    expect(response.session).to.have.property('refresh_token');
+    expect(response.session.refresh_token).to.not.be.empty;
+  }).timeout(10000);
+
+  it('should be refreshed if access token is expired', async function () {
```
mobitar, line 98: This is a good test 👍 Another thing we should probably test for is expecting that the new token expireAt > Date.now().

test/session.test.js
mobitar: (I don't necessarily mean you should just have one `expect` call per test, but just that we shouldn't repeat expect calls across multiple tests. And generally the less expect calls per test, the better)

test/session.test.js
```diff
@@ -0,0 +1,160 @@
...
+    expect(sessionBeforeSync.accessToken).to.not.equal(sessionAfterSync.accessToken);
+    expect(sessionBeforeSync.refreshToken).to.not.equal(sessionAfterSync.refreshToken);
+    expect(sessionBeforeSync.expireAt).to.be.lessThan(sessionAfterSync.expireAt);
+  }).timeout(20000);
+
+  it('should be consistent between storage and apiService', async function () {
+    const sessionFromStorage = await getSessionFromStorage(this.application);
+    const sessionFromApiService = this.application.apiService.session;
+
+    expect(sessionFromStorage).to.equal(sessionFromApiService);
```
mobitar, line 122: How is it possible that a raw session from storage is equal to a Session object 🤔The former should be a generic JS object, and the latter would be a Session instance.

test/session.test.js
johnny243: Currently, the `syncing-server` is checking token expiration to the second: 

`session.rb`
```ruby
def expired_access_token?
    access_token_expire_at < DateTime.now.to_i
end
```

`DateTime.now.to_i` returns a 10-digit integer (in seconds).

So we may need to rewrite it to use a 13-digit integer (with milliseconds) in order to be more precise.

mobitar:
Some more tests we should have:

- Signing out with valid token
- Changing password with valid token
- Changing password with invalid token
- Syncing while a refresh token request is in progress
- Signing out and signing back in, and ensuring tokens are properly set up
- Renewing a session with an expired renewal token
- Renewing a session with an invalid renewal token (incorrect/unauthorized/corrupted value)
- Attempting to use a session token after sign out should return unauthorized

Also, we need to introduce scenarios with regards to saving data when a renewal is required:
- Saving a note and running sync with an expired token should have expected synced values after sync is complete. (That is, await sync() should also await refreshToken(), so when the original sync is complete, we should expect that the note values have properly been saved. We can do this by signing out and signing back in to download data, and ensure the downloaded data values match what we expect).

test/session.test.js
mobitar: Hmm, might be good to be more precise than. `to_f` instead of `to_i` perhaps. What are the pros/cons?

johnny243: Let's use `DateTime.now.strftime('%Q').to_i`.

It returns a timestamp with milliseconds.

test/session.test.js
johnny243: In this case, `status` is a property that is returned in the response. I'm checking that it has an specific value when the session is refreshed successfully. I will need to repeat this `expect` call on multiple tests. Perhaps I should access the property directly, instead of checking that it exists first? It should look like:

```javascript
it('should return the new session in the response when refreshed', async function () {
    const response = await this.application.apiService.refreshSession();

    expect(response.status).to.equal(200);
    expect(response.token).to.be.a('string');
    expect(response.token).to.not.be.empty;
    expect(response.session.expire_at).to.be.a('number');
    expect(response.session.refresh_token).to.not.be.empty;
  }).timeout(10000);
```

test/session.test.js
johnny243: The empty array is for the `payloads` param: https://github.com/standardnotes/snjs/blob/fbc5ec68ad47c47eaccf67ceb7971b0b43f8b36a/lib/services/api/api_service.ts#L231

If the `payloads` param is not provided, [this line](https://github.com/standardnotes/snjs/blob/fbc5ec68ad47c47eaccf67ceb7971b0b43f8b36a/lib/services/api/api_service.ts#L244) will fail because of the `map` method running on an undefined variable.

Also, I'm using `this.application.apiService.sync([])` and not `this.application.syncService.sync()` because I want to have the response available.

test/session.test.js
johnny243: Good question. I sort of missed that detail there. Somehow, the session from `storageService` is a `Session` object. But it is not clear to my why...

test/session.test.js
mobitar: Ohh I didn't catch that this was the apiService and not syncService.

johnny243:
- [x] Signing out with valid token
- [x] Changing password with valid token
- [x] Changing password with invalid token
- [ ] Syncing while a refresh token request is in progress
- [x] Signing out and signing back in, and ensuring tokens are properly set up
- [ ] Renewing a session with an expired renewal token
- [ ] Renewing a session with an invalid renewal token (incorrect/unauthorized/corrupted value)
- [x] Attempting to use a session token after sign out should return unauthorized

lib/services/api/api_service.ts
```diff
@@ -188,7 +188,9 @@ export class SNApiService extends PureService {
       url,
       undefined,
       this.session!.accessToken
-    );
+    ).then(async (response) => {
```
mobitar, line 191: This is redundant, as the response will be what is returned by the return statement on 187. This would only be useful if you want to modify response before returning it or if you're wanting to return a subset of the response object.

johnny243: My bad, I was debugging the response object and forgot to remove this :sweat_smile:

lib/services/api/messages.ts
```diff
@@ -14,8 +14,8 @@ export const API_MESSAGE_FALLBACK_LOGIN_FAIL       = 'Invalid email or password.
 export const API_MESSAGE_GENERIC_TOKEN_REFRESH_FAIL= `A server error occurred while trying to refresh your session.
                                                       Please try again.`;
 
-export const API_MESSAGE_TOKEN_REFRESH_IN_PROGRESS = `Your account session is being renewed with the server.
-                                                      Please try your request again.`;
+export const API_MESSAGE_TOKEN_REFRESH_IN_PROGRESS = `Your account session is being renewed with the server. \
```
johnny243, line 17: @mobitar I had to modify this string a bit. I wanted to test the message in https://github.com/standardnotes/snjs/commit/824c6d6f90b9dba5072ce86598f4f56bcf739627#diff-3625700b711b80a59b096f9544c87c27R262 but it wouldn't let me do so (unless that I copied the exact string with spaces, which doesn't sound to be a good approach). 

Any comments on this matter? Should we use this string format for the rest of messages?

test/session.test.js
```diff
@@ -0,0 +1,265 @@
...
+    expect(syncResponse.error.message).to.equal('Invalid login credentials.');
+  }).timeout(20000);
+
+  describe('change password request', async function () {
+    beforeEach(async function () {
+      this.newPassword = Factory.randomString();
+    });
+
+    it('should be successful with a valid access token', async function () {
+      const changePasswordResponse = await this.application.changePassword(
```
johnny243, line 138: When running `this.application.changePassword()`, the token will expire before the request is sent to the server (using an expiration time of 2 seconds). A quick workaround would be to use a higher expiration time (but that just increases the overall time the tests are executing). 

I guess this is normal, since a new set of keys are generated.

lib/services/api/messages.ts
mobitar: We don't actually want newlines in these messages. We should instead use "" + "" syntax for multi-line strings. It was a mistaken assumption on my end that using `` syntax would automatically ignore newlines in the JS files.

mobitar:
- [ ] Test for limiting recursion of failed renewal requests

PR #12, opened 2020-05-13
fix: itemsKeys bug

baptiste-grob:
Compare against an up to date itemsKeys value after running operations in the itemManager

### DIFF
--- a/lib/application.ts
+++ b/lib/application.ts
@@ -851,7 +851,7 @@ export class SNApplication {
    * https://github.com/standardnotes/desktop/issues/131
    */
   private async rewriteItemsKeys() {
-    const itemsKeys = this.protocolService!.itemsKeys();
+    const itemsKeys = this.itemManager!.itemsKeys();
     const payloads = itemsKeys.map((key) => key.payloadRepresentation());
     await this.storageService!.deletePayloads(payloads);
     await this.syncService!.persistPayloads(payloads);
--- a/lib/application.ts
+++ b/lib/application.ts
@@ -121,11 +121,11 @@ export class SNApplication {
    * @param platform The Platform that identifies your application.
    * @param namespace A unique identifier to namespace storage and
    *  other persistent properties. Defaults to empty string.
-   * @param crypto The platform-dependent instance of SNCrypto to use. 
+   * @param crypto The platform-dependent instance of SNCrypto to use.
    * Web uses SNWebCrypto, mobile uses SNReactNativeCrypto.
-   * @param swapClasses Gives consumers the ability to provide their own custom 
-   * subclass for a service. swapClasses should be an array  of key/value pairs 
-   * consisting of keys 'swap' and 'with'.  'swap' is the base class you wish to replace, 
+   * @param swapClasses Gives consumers the ability to provide their own custom
+   * subclass for a service. swapClasses should be an array  of key/value pairs
+   * consisting of keys 'swap' and 'with'.  'swap' is the base class you wish to replace,
    * and 'with'  is the custom subclass to use.
    * @param skipClasses An array of classes to skip making services for.
    */
@@ -186,8 +186,8 @@ export class SNApplication {
    * Runs migrations, handles device authentication, unlocks application, and
    * issues a callback if a device activation requires user input
    * (i.e local passcode or fingerprint).
-   * @param awaitDatabaseLoad  
-   * Option to await database load before marking the app as ready. 
+   * @param awaitDatabaseLoad
+   * Option to await database load before marking the app as ready.
    */
   public async launch(awaitDatabaseLoad = false) {
     this.launched = false;
@@ -254,7 +254,7 @@ export class SNApplication {
     }, DEFAULT_AUTO_SYNC_INTERVAL);
   }
 
-  /** 
+  /**
    * The migrations service is initialized with this function, so that it can retrieve
    * raw challenge values as necessary.
    */
@@ -278,7 +278,7 @@ export class SNApplication {
     }
   }
 
-  /** 
+  /**
    * @param singleEvent Whether to only listen for a particular event.
    */
   public addEventObserver(
@@ -315,7 +315,7 @@ export class SNApplication {
     this.migrationService!.handleApplicationEvent(event);
   }
 
-  /** 
+  /**
    * Whether the local database has completed loading local items.
    */
   public isDatabaseLoaded() {
@@ -337,28 +337,28 @@ export class SNApplication {
     await this.syncService!.sync();
   }
 
-  /** 
+  /**
    * Finds an item by UUID.
    */
   public findItem(uuid: string) {
     return this.itemManager!.findItem(uuid);
   }
 
-  /** 
+  /**
    * Returns all items.
    */
   public allItems() {
     return this.itemManager!.items;
   }
 
-  /** 
+  /**
    * Finds an item by predicate.
   */
   public findItems(predicate: SNPredicate) {
     return this.itemManager!.itemsMatchingPredicate(predicate);
   }
 
-  /** 
+  /**
    * Finds an item by predicate.
    */
   public getAll(uuids: UuidString[]) {
@@ -373,7 +373,7 @@ export class SNApplication {
     return this.itemManager!.emitItemFromPayload(item.payloadRepresentation(), source);
   }
 
-  /** 
+  /**
    * Creates a managed item.
    * @param needsSync  Whether to mark the item as needing sync. `add` must also be true.
    */
@@ -392,7 +392,7 @@ export class SNApplication {
     return item;
   }
 
-  /** 
+  /**
    * Creates an unmanaged item that can be added later.
    * @param needsSync  Whether to mark the item as needing sync. `add` must also be true.
    */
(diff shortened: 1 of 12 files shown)


lib/services/protocol_service.ts
```diff
@@ -1150,21 +1150,21 @@ export class SNProtocolService extends PureService implements EncryptionDelegate
...
    */
   private async handleDownloadFirstSyncCompletion() {
     /**
     * Find items keys with null or epoch updated_at value, indicating
     * that they haven't been synced yet.
     */
-    const itemsKeys = this.itemsKeys;
-    const neverSynced = itemsKeys.filter((key) => {
+    const itemsKeys = this.itemsKeys();
+    const neverSyncedKey = itemsKeys.filter((key) => {
```
mobitar, line 1167: This should be pluralized.

lib/services/protocol_service.ts
```diff
@@ -1150,21 +1150,21 @@ export class SNProtocolService extends PureService implements EncryptionDelegate
...
    * delete any never-synced items keys we have here locally.
    */
   private async handleDownloadFirstSyncCompletion() {
     /**
     * Find items keys with null or epoch updated_at value, indicating
     * that they haven't been synced yet.
     */
-    const itemsKeys = this.itemsKeys;
-    const neverSynced = itemsKeys.filter((key) => {
+    const itemsKeys = this.itemsKeys();
```
mobitar, line 1166: We probably don't need to make a variable out of this, as it's just being used once. This way would make it more consistent with how we access `this.itemsKeys()` inline below.

mobitar:
There are 3 unit tests that still access `itemManager.itemsKeys`.

Also protocolService still uses getter style. Should probably be changed to function style as well. (Don't forget about unit tests as well).

You should be sure that all tests pass before every PR.

lib/services/protocol_service.ts
baptiste-grob: It's used a second time on line 1174:
```
const defaultSyncedKey = itemsKeys.find((key) => { /* ... */ }
```

mobitar: Someone maintaining this code in the future might see the second this.itemsKeys() reference and think, oh, they forget they created a variable, and they'll swap out the function call for the variable. How do we prevent that 😅

mobitar: (Also no tests in place that would catch that regression)

lib/utils.ts
```diff
@@ -159,11 +159,15 @@ export function existsInArray<T>(inArray: T[], value: T) {
...
 }
 
-/**
- * Removes an object from the array by reference equality or shallow value equality
+ * Removes an object from the array by strict equality
  */
 export function removeFromArray<T>(array: T[], value: T) {
-  array.splice(array.indexOf(value), 1);
+  const valueIndex = array.indexOf(value);
+  if (valueIndex === -1) {
```
mobitar, line 160: Hmm..this is more a matter of design opinion, but the warn here seems odd. We should just instead describe the expected behavior in the method comment. "If the value does not exist in the array, no action is taken." In my experience you either throw an exception here (Python) or do a no-op (most other languages). The in between of a warn where a human has to be paying attention to catch it seems too arbitrary. What do you think?

baptiste-grob: Yes, I wasn't very enthused by this solution either but I lacked the domain knowledge to make another decision at the time. I've seen warnings pop up in the console while I was running tests, which made me see that this situation could sometimes present itself. What I'm wondering is how much of it is intentional… Maybe this calls for two different functions, a strict one for when you know the element to be in the array and a soft one for when you don't?

mobitar: Warnings in the console written by me tend to follow one of these patterns:

- You're accessing a deprecated function
- You're using a function incorrectly but not in a way that would jeopardize behavior

But in this case as long as we document that it is a no-op if the element doesn't exist, then it is not being used incorrectly. I think it's safe to go without this warning in that case.

lib/utils.ts
```diff
@@ -142,50 +142,48 @@ export function extendArray<T>(inArray: T[], otherArray: T[]) {
...
-/** 
- * Removes all items appearing in toSubtract from inArray, in-place 
+/**
+ * Removes all items appearing in toSubtract from inArray, in-place
  * @param toSubtract - The list of items to remove from inArray
  */
 export function subtractFromArray<T>(inArray: T[], toSubtract: T[]) {
   for (const value of toSubtract) {
-    inArray.splice(inArray.indexOf(value), 1);
+    removeFromArray(inArray, value);
```
baptiste-grob, line 151: If `value` is not in `inArray` this would execure `inArray.splice(-1,1)`, removing the _last_ element in the array.

lib/utils.ts
```diff
@@ -142,50 +142,48 @@ export function extendArray<T>(inArray: T[], otherArray: T[]) {
...
   }
 }
 
-/** 
- * Determines if value exists in array, by reference
- */
-export function existsInArray<T>(inArray: T[], value: T) {
-  return inArray.indexOf(value) >= 0;
-}
-
```
baptiste-grob, line 161: We can use Array.prototype.includes instead for the same behavior

lib/utils.ts
baptiste-grob: I pushed some new changes to align with your recommendation, I'll let you resolve the conversation if you think it's good now :)

test/protocol.test.js
```diff
@@ -76,13 +76,6 @@ describe('protocol', function () {
     expect(result.errorDecrypting).to.not.be.ok;
   });
 
-  it('decrypting already decrypted payload should return same payload', async function () {
-    const payload = Factory.createNotePayload();
-    const result = await this.application.protocolService.payloadByDecryptingPayload(payload);
-    expect(payload).to.equal(result);
-    expect(result.errorDecrypting).to.not.be.ok;
-  });
-
```
baptiste-grob, line 85: Duplicate test, see line 72

test/utils.test.js
```diff
@@ -48,10 +48,35 @@ describe('utils', () => {
...
+      expect(array).to.eql([1, 3]);
+    });
+
+    it('Ignores items not appearing in the array', () => {
+      const array = [1, 2, 3];
+      removeFromArray(array, 0);
+      expect(array).to.eql([1, 2, 3]);
+      removeFromArray(array, {});
+    });
   });
```
baptiste-grob, line 80: New tests to ensure our array removal functions don't exhibit undefined behavior.

test/protocol.test.js
baptiste-grob, line 85: Duplicate test

baptiste-grob, line 85: Duplicate test, see line 72

lib/utils.ts
mobitar: 😵Good point! I didn't catch this.

lib/services/protocol_service.ts
baptiste-grob: I've added a regression test: https://github.com/standardnotes/snjs/pull/12/commits/1f72faa785d896287263d17ec5e3657067bbaf1f

lib/application.ts
```diff
@@ -1021,19 +1021,22 @@ export class SNApplication {
       const syncPromise = this.syncService!.sync({
         mode: SyncModes.DownloadFirst,
         checkIntegrity: true,
-        queueStrategy: SyncQueueStrategy.ForceSpawnNew
+        queueStrategy: SyncQueueStrategy.ForceSpawnNew,
+        awaitAll: true,
```
mobitar, line 1025: I thought this should be `awaitAll: awaitSync,`? You don't want to awaitAll on regular sign in because it will completely block the UI until all X user notes are downloaded (could be 3000).

lib/services/item_manager.ts
```diff
@@ -142,7 +142,7 @@ export class ItemManager extends PureService {
   /**
    * Returns all non-deleted items keys
    */
-  get itemsKeys() {
+  itemsKeys() {
```
mobitar, line 145: Did we decide not to rename this `latestItemsKeys`?

baptiste-grob: We made that decision for the `protocol_service.ts` method: https://github.com/standardnotes/snjs/pull/12/commits/efaff05eac248c7c84f83a3fbb532dc0a336914a
Because it was to cover one specific case that was internal to that class, which is also why I decided to make the method private. I don't feel like it would make much sense to a caller when used as a public method on the item manager.

mobitar: Would it be less confusing to just remove one of the implementations so there is only one source of itemsKeys? I think maybe protocol service should remove its function.

baptiste-grob: Well, I like when there is less indirection going on. However the reason we went with `latestItemsKeys()` was to provide some documentation in reaction to the bug we experienced. Still, you could argue that it wouldn't have prevented the bug anyway.
So I'm in favor of removing it but I don't feel very strongly about the decision. Can I let you make the call here?

mobitar: It definitely is subtle. But it seems like really it should be the public function that should be renamed, not the private one, since public consumption is more liable to misuse. So if it was instead itemManager.itemsKeys that was renamed to latestItemsKeys, that would seem to make more sense. Then the private implementation in protocol service can either be removed or renamed to match the itemsManager name, this way we know 100% that they are the exact same function.

baptiste-grob: Okay, I will rename the public function then. This poses the question of what to do with the other functions in itemManager: https://github.com/standardnotes/snjs/pull/12/files#diff-b92d581b5582e43bac311c8c05b33eb0R152-R168
Should they all be renamed to latestNotes, latestTags, etc?

PR #14, opened 2020-05-29
feat: ability to provide flags to regex

johnny243:
Specially useful to create SmartTags with case insensitive match.

Example syntax:

```
!["Daylog entries", "title", "matches", ["^daylog", "gi"]]
```

### DIFF
--- a/lib/models/core/predicate.ts
+++ b/lib/models/core/predicate.ts
@@ -158,7 +158,10 @@ export class SNPredicate {
       return this.resolveIncludesPredicate(valueAtKeyPath, targetValue);
     }
     else if (predicate.operator === 'matches') {
-      const regex = new RegExp(targetValue as string);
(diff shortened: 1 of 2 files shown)


johnny243:
@mobitar will mark this PR as ready once `004` is merged into `master` branch.

mobitar:
Hmm, what if the flags were part of the main string, rather than introducing an array structure?

johnny243:
I think that would be possible if we had static pattern enclosed in forward slashes:

```javascript
const regexp = /^daylog/gi;
regexp.test("Daylog"); // true
``` 

But since we are receiving a dynamic pattern from our users, we need to use the `RegExp` object, and provide both the pattern and flags separately.

karolsojko:
Please rebase this PR onto to the current default branch which is `master`. Sorry for the trouble.

PR #15, opened 2020-06-04
Remove sncrypto dependency

radko93:
This removes sncrypto from snjs, as it breaks mobile and adds a lot of unnecessary code to snjs bundle. It's still there as devDependency to make tests pass. Clients now have to pass `SNPureCrypto` implementation. This will, unfortunately, require web to change as well and install `sncrypto` as its own dependency.


radko93:
As agreed with @mobitar, we will first make `sncrypto-interface` as a separate package. @mobitar you can use SNPureCrypto from this PR to do this.

mobitar:
Heads up @baptiste-grob. We've refactored sncrypto so that it doesn't import libsodium needlessly (mobile doesn't use it).
- snjs imports sncrypto but only requires sncrypto/sncrypto-common to get a reference to the SNPureCrypto interface.
- snjs uses sncrypto/sncrypto-web only in unit tests.

https://github.com/standardnotes/sncrypto/pull/2/files

radko93:
There's a problem with this PR, `SNPureCrypto` should be imported from `sncrypto` but I cannot find a way to do that right now.

PR #18, opened 2020-06-09
feat: tests for actions

johnny243:
Tests for `actionsService`.


lib/services/actions_service.ts
```diff
@@ -107,7 +107,7 @@ export class SNActionsService extends PureService {
     return this.httpService!.getAbsolute(
       extension.url,
       params
-    ).then((response) => {
+    ).then(async (response) => {
```
baptiste-grob, line 110: Since you're already inside an async function it might be clearer/more consistent to use the await syntax here, i-e:
```
const response = await this.httpService.getAbsolute(…)
```

johnny243: I'm not following, sorry :sweat_smile: . The line you're pointing is the `then` callback, how would I use `await` as you say here?

I added `async` here just so I can `await` inside the `then` callback:

 ``` javascript
this.itemManager!.changeActionsExtension(...)
```

Comment (mobitar) [reply lib/services/actions_service.ts:110]: Basically instead of

```javascript
return this.httpService!.getAbsolute(
      extension.url,
      params
    ).then((response) => {
```

You would do

```javascript
const response = this.httpService!.getAbsolute(
      extension.url,
      params
)
```

which ins indeed cleaner.

PR #24, opened 2020-06-23
Allow reimporting deleted notes

baptiste-grob:
Lets a user to restore their previously-deleted notes from a backup

### DIFF
--- a/lib/protocol/payloads/deltas/remote_conflicts.ts
+++ b/lib/protocol/payloads/deltas/remote_conflicts.ts
@@ -31,18 +31,14 @@ export class DeltaRemoteConflicts extends PayloadsDelta {
         payload.uuid!,
         PayloadSource.DecryptedTransient
       );
-      if (!decrypted) {
+      if (!decrypted && !payload.deleted) {
         /** Decrypted should only be missing in case of deleted payload */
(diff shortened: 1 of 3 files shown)


lib/protocol/payloads/deltas/remote_conflicts.ts
```diff
@@ -31,18 +31,14 @@ export class DeltaRemoteConflicts extends PayloadsDelta {
         payload.uuid!,
         PayloadSource.DecryptedTransient
       );
-      if (!decrypted) {
+      if (!decrypted && !payload.deleted) {
         /** Decrypted should only be missing in case of deleted payload */
-        if (!payload.deleted) {
-          throw 'Unable to find decrypted counterpart for data conflict.';
-        }
-        results.push(payload);
```
mobitar, line 39: Just want to make sure this line wasn't overlooked when simplifying the `if` statement? I assume the pushing here is handled instead by passing this payload to the conflict delta.

baptiste-grob: Exactly. This seemed like a responsibility for the `ConflictDeta` class and not its callers. The test would have covered that oversight as well, however your comment made me think of an way to make said test tighter.

PR #25, opened 2020-06-25
fix: correct WebPrefKey for EditorMonospaceEnabled

johnny243:
The enum value for `EditorMonospaceEnabled` should be [monospaceEnabled](https://github.com/standardnotes/web/blob/4dbf879f873cfea56bb6e8f9599bcf6453a60961/app/assets/javascripts/controllers/editor.js#L806).

### DIFF
--- a/lib/models/app/userPrefs.ts
+++ b/lib/models/app/userPrefs.ts
@@ -6,7 +6,7 @@ export enum WebPrefKey {
   NotesPanelWidth = 'notesPanelWidth',
   EditorWidth = 'editorWidth',
   EditorLeft = 'editorLeft',
-  EditorMonospaceEnabled = 'monospaceFont',
+  EditorMonospaceEnabled = 'monospaceEnabled',
   EditorSpellcheck = 'spellcheck',
   EditorResizersEnabled = 'marginResizersEnabled',
   SortNotesBy = 'sortBy',


johnny243:
> Where is this string used? Will changing this essentially reset the setting for every note for existing users?

See [here](https://github.com/standardnotes/web/blob/cd43234b5e8d23fff9e00c4958e68467f6868660/app/assets/javascripts/services/preferencesManager.js#L7). On `master` we are still using `monospaceFont`, so I guess we should stick with it in order to preserve user preferences.

mobitar:
Wait what..I don't understand this change. What's incorrect about the current value? Just stylistically? If we do that we'd have to migrate all users current key to the new one.

johnny243:
> If we do that we'd have to migrate all users current key to the new one.

My apologies, I tested this with newly created accounts only.

Basically, the issue comes from [this line](https://github.com/standardnotes/web/blob/48261087c1e4b7acd624ff88bd20842059a8d70d/app/assets/javascripts/views/editor/editor_view.ts#L968), which is setting the state value for `monospaceEnabled`, which does not exist in `WebPrefKey`.

We should instead use `monospaceFont` in this line and in other places that may reference it. This will prevent us from migrating existing users to a new key.

What do you think?

PR #27, opened 2020-06-25
Fix: upgrade with no account and a passcode

baptiste-grob:
I investigated doing this during the storage upgrade but it would have involved much more work, so instead it works like account-based upgrades.

### DIFF
--- a/lib/services/storage_service.ts
+++ b/lib/services/storage_service.ts
@@ -178,7 +178,7 @@ export class SNStorageService extends PureService {
     const wrappedValue = this.values[ValueModesKeys.Wrapped];
     const decryptedPayload = await this.decryptWrappedValue(wrappedValue);
     if (decryptedPayload.errorDecrypting) {
-      throw 'Unable to decrypt storage.';
+      throw Error('Unable to decrypt storage.');
     }
     this.values[ValueModesKeys.Unwrapped] = Copy(decryptedPayload.contentObject);
     delete this.values[ValueModesKeys.Wrapped];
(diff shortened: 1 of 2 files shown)


lib/services/storage_service.ts
```diff
@@ -135,7 +138,11 @@ export class SNStorageService extends PureService {
   }
 
   public async canDecryptWithKey(key: SNRootKey) {
-    const wrappedValue = this.values[ValueModesKeys.Wrapped];
+    let wrappedValue = this.values[ValueModesKeys.Wrapped];
+    if (!wrappedValue) {
+      const values = await this.getValuesFromStorage();
```
mobitar, line 143: Why would `wrappedValue` be null here (other than it being truly null because it doesn't exist in storage)? `initializeFromDisk` is called way before `canDecryptWithKey` is, so calling  `getValuesFromStorage` in here should not have any results different from what would have been set in place by `initializeFromDisk `.

baptiste-grob: When we decrypt the storage, we delete `this.values[ValueModesKeys.Wrapped]`. I've not investigated why so maybe you can enlighten me here. Is it for security?

mobitar: Ah I see. I was evaluating this from a launch context. We delete that value because it was seemingly no longer needed for normal app functioning. But in this case if you attempt to verify passcode in the middle of app lifecycle, then it would indeed be missing. In this case I believe the better solution would be to remove `delete this.values[ValueModesKeys.Wrapped];` from `decryptStorage`. Because we add this value back in `repersistToDisk` anyway, which is called many times during normal app lifecycle.

baptiste-grob: Gotcha, I've pushed an update.

PR #29, opened 2020-06-29
fix: allow clearing storage when app is locked

baptiste-grob:
We need this to let the user clear their storage when they can't remember their passcode

### DIFF
--- a/lib/services/storage_service.ts
+++ b/lib/services/storage_service.ts
@@ -176,11 +176,35 @@ export class SNStorageService extends PureService {
     this.values[ValueModesKeys.Unwrapped] = Copy(decryptedPayload.contentObject);
   }
 
+  /** @todo This function should be debounced. */
+  private async persistValuesToDisk() {
+    if (!this.storagePersistable) {
+      return;
+    }
+    if (this.persistencePolicy === StoragePersistencePolicies.Ephemeral) {
+      return;
+    }
+    const values = await this.immediatelyPersistValuesToDisk();
+    /** Save the persisted value so we have access to it in memory (for unit tests afawk) */
+    this.values[ValueModesKeys.Wrapped] = values[ValueModesKeys.Wrapped];
+  }
+
+  private async immediatelyPersistValuesToDisk(): Promise<StorageValuesObject> {
+    return this.executeCriticalFunction(async () => {
+      const values = await this.generatePersistableValues();
+      await this.deviceInterface!.setRawStorageValue(
+        this.getPersistenceKey(),
+        JSON.stringify(values)
+      );
+      return values;
+    });
+  }
+
   /**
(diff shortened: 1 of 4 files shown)


mobitar:
Hmm..I'd classify this as a hack. What about adding a `force` parameter to `repersistToDisk` with documentation regarding its purpose?

lib/services/storage_service.ts
```diff
@@ -203,12 +203,14 @@ export class SNStorageService extends PureService {
...
-      return;
-    }
-    if (this.persistencePolicy === StoragePersistencePolicies.Ephemeral) {
-      return;
+  private async repersistToDisk(force = false) {
+    if (!force) {
+      if (!this.storagePersistable) {
+        return;
+      }
+      if (this.persistencePolicy === StoragePersistencePolicies.Ephemeral) {
```
mobitar, line 211: Should `force` encapsulate the `persistencePolicy` as well? I'd say no. If ephemeral we should never write to storage. In this case the less power `force` has the better. At least until we explicitly need this power. For now we only need it to override `storagePersistable`.

You should definitely document/comment somewhere why we might need to use force here.

baptiste-grob:
In that case I feel like this calls for creating a separate function to sort of corner that complexity away from the most common behavior. But if you prefer a boolean given the current complexity of the function I wouldn't mind that much.

mobitar:
Yeah definitely. That was just one idea. I'm not a huge fan of the `force` approach either but I thought it would be more explicit than temporarily setting a bool.

baptiste-grob:
Yes, you're definitely right. Let me get back with a better approach.

lib/services/storage_service.ts
```diff
@@ -293,7 +296,7 @@ export class SNStorageService extends PureService {
    */
   async clearValues() {
     this.setInitialValues();
-    await this.repersistToDisk();
+    await this.immediatelyPersistValuesToDisk();
```
mobitar, line 299: The one difference I can find between the old and new implementations is that this new placement in `clearValues` does not execute the persistence in a `executeCriticalFunction` block. I can't recall the specifics of timing here, but to be safe, we should perhaps move the `executeCriticalFunction` call from `persistValuesToDisk` to `immediatelyPersistValuesToDisk`. This way, we know it's always executed as a critical function.

baptiste-grob: Good point. I've pushed changes to address this.
Let me know if you have a better idea for the method's name, I've really been struggling with this one.

mobitar: I think it's fine. Typically with this pattern I go with persistValuesToDisk/performPersistValuesToDisk. But in this case since we have a todo to debounce the former function, the "immediately" makes sense.

PR #30, opened 2020-06-29
fix: clear challenge operations on deinit

baptiste-grob:


mobitar:
Btw, I find it acceptable to do `(this.challengeOperations as any) = undefined` inside of `deinit`, rather than modifying the property definition to be optional, for example, or to create a new object. I believe we do this in many places.

I'm curious what this fixes btw? Weird that this value would be depended on during or post deinit.

baptiste-grob:
I think I made a false assumption. My initial thought was that we should clear existing challenges that clients didn't respond to on app deinit. But since we clear the challenge service field, there should be no need for this.

mobitar:
Ah ok. For the record essentially the point of deinit is to ensure we don't have any circular references with objects and services. And to clear out any sensitive values just to ensure they don't linger. But if an object is properly dealloced then any sensitive values should be cleared anyway.

PR #31, opened 2020-06-29
feature: expose encryption policy value

radko93:
Some types change are related to this PR.


radko93:
@mobitar please check

PR #34, opened 2020-06-30
fix: importing with encryptedPayloads

johnny243:
This PR solves the following issues:

## Error decrypting itemKey parameters

When importing an encrypted backup file with version `004`, you need to compute the `SNRootKey` (using the `password`) to decrypt your `SNItemsKey` items, but you don't need such `SNRootKey` to decrypt the rest of your items. The respective `SNItemsKey` will be used for this purpose. 

When trying to decrypt a `SNNote` item for example, `payloadByDecryptingPayload` [will have a root key passed as the key param](https://github.com/standardnotes/snjs/blob/c1b934fca11e4acc6f2d8d62d3786eca15dbcf02/lib/services/protocol_service.ts#L516), and not an `SNItemsKey`. [This line](https://github.com/standardnotes/snjs/blob/c1b934fca11e4acc6f2d8d62d3786eca15dbcf02/lib/protocol/operator/004/operator_004.ts#L201) will return a `null` value, and thus, output to console the following error when importing an encrypted backup file with version `004`:

```
Error decrypting itemKey parameters
```

**Solution**: https://github.com/standardnotes/snjs/pull/34/commits/da48c9304bab0142868ae7031670a052454c8a71

## Payloads that are waitingForKey are not decrypted

It's simply because such `payloads` are not yet valid:

https://github.com/standardnotes/snjs/blob/316b1b0caa00bb2ad5dde07b3a473c1dba2a65c7/lib/application.ts#L768

And will not be included in the above array. So they must be imported separately, before or after importing valid `payloads`.

Basically, this is how they will be decrypted:

1. https://github.com/standardnotes/snjs/blob/316b1b0caa00bb2ad5dde07b3a473c1dba2a65c7/lib/application.ts#L764 this line will return all `payloads` (including those that are still `waitingForKey`).
2. At this point, we can import `payloads` that are `waitingForKey = true`
3. After the above, we import valid `payloads`
4. The `ItemsKey` observer will trigger `decryptErroredItems()`: https://github.com/standardnotes/snjs/blob/316b1b0caa00bb2ad5dde07b3a473c1dba2a65c7/lib/services/protocol_service.ts#L146
5. `decryptErroredItems()` will find `invalidItems` from the `itemManager` (this will get `payloads` that were previously imported on step 2)
6. The `payloads` will be decrypted by finding the corresponding `ItemsKey`: https://github.com/standardnotes/snjs/blob/316b1b0caa00bb2ad5dde07b3a473c1dba2a65c7/lib/services/protocol_service.ts#L526

**Solution(s)**: https://github.com/standardnotes/snjs/pull/34/commits/574f99b896475863297bc99ea3ff9e116530e7e5, https://github.com/standardnotes/snjs/pull/34/commits/9be44e2172986fdf2367c0c062141259b1f3d4bc and https://github.com/standardnotes/snjs/pull/34/commits/6c0191e41b24e787dd6c5cbf3b16868e754c1ccc

---

### Side notes

- It is best to first import valid `payloads` and if any where imported, we can proceed importing `payloads` that are `waitingForKey`. This is good for the case when an invalid password was provided. So we avoid importing any `payloads` that are `waitingForKey` that will never have a key in this particular case. 

The above was addressed in https://github.com/standardnotes/snjs/pull/34/commits/d9e72c82ecd7b43de6d8aa02efa5277fd303c724.


test/model_tests/importing.test.js
```diff
@@ -270,4 +270,37 @@ describe('importing', () => {
       expect(this.application.itemManager.tags.length).to.equal(1);
       expect(this.application.findItem(note.uuid).deleted).to.be.false;
     });
+
+    it('should import data from encrypted payload', async function () {
```
mobitar, line 274: This test should be called "should import data from 003 encrypted payload" and we should have another test called "should import data from 004 encrypted payload"

lib/services/protocol_service.ts
```diff
@@ -612,7 +611,9 @@ export class SNProtocolService extends PureService implements EncryptionDelegate
       );
     });
     let decryptedPayloads;
-    if (keyParams) {
+    let keyParams;
+    if (keyParams = data.keyParams || data.auth_params) {
```
mobitar, line 615: Hmm..don't really like the style of this. It may work but it's not a pattern we use anywhere else in the lib.

What about

```javascript
const keyParamsData = data.keyParams || data.auth_params;
if(keyParamsData) {
   const keyParams = this.createKeyParams(keyParamsData);
}
```

lib/services/protocol_service.ts
```diff
@@ -572,11 +572,30 @@ export class SNProtocolService extends PureService implements EncryptionDelegate
...
           encryptedPayload,
-          key
+          keyForPayload
         );
-        decryptedPayloads.push(decryptedPayload);
+        /**
+         * If waitingForKey, we must import the payload in order to decrypt it later.
+         */
+        if (decryptedPayload.waitingForKey) {
+          await this.modelManager!.importPayloads([decryptedPayload]);
```
johnny243, line 595: @mobitar if a `Payload` has `waitingForKey = true` it must be imported into the items record:
https://github.com/standardnotes/snjs/blob/a27979314775f7aa34d4817bee740ab3fa3948c5/lib/services/protocol_service.ts#L601
The above line won't return anything because such items do not belong to the items record yet.

johnny243: https://github.com/standardnotes/snjs/blob/a27979314775f7aa34d4817bee740ab3fa3948c5/lib/application.ts#L787

The above will import all _valid items_ (i.e: `ItemsKey`)

johnny243: And then, the `ItemsKey` observer will have such `ItemsKey`s and will call [decryptErroredItems()](https://github.com/standardnotes/snjs/blob/a27979314775f7aa34d4817bee740ab3fa3948c5/lib/services/protocol_service.ts#L600) to attempt to decrypt invalid items (i.e: items which are `waitingForkey = true`):

https://github.com/standardnotes/snjs/blob/a27979314775f7aa34d4817bee740ab3fa3948c5/lib/services/protocol_service.ts#L143

mobitar, line 595: This is no good. This function should not have any side effects. It should be functional. Pretend it is a utility method. It should not trigger state updates.

mobitar: (This is regarding `await this.modelManager!.importPayloads([decryptedPayload]);`. Should not be doing this here.)

lib/services/protocol_service.ts
```diff
@@ -572,11 +572,30 @@ export class SNProtocolService extends PureService implements EncryptionDelegate
         continue;
       }
       try {
+        let keyForPayload = key;
+        const version = encryptedPayload!.version;
+        const payloadContentType = encryptedPayload!.content_type;
+        const keyContentType = key?.content_type;
+        /** We check if the key can decrypt the encrypted payload. */
+        if (
+          version === ProtocolVersion.V004 &&
```
mobitar, line 581: This needs to be more future proof. Imagine that V005 comes out one month later with a minor change. Do we keep adding if conditions here for every new version?

lib/services/protocol_service.ts
johnny243: What about inside `importData`: https://github.com/standardnotes/snjs/blob/a27979314775f7aa34d4817bee740ab3fa3948c5/lib/application.ts#L759

We can import items there before importing valid items.

mobitar: That would seem to make sense, but I'd have to see the solution in its final form to see how it plays out.

lib/application.ts
```diff
@@ -765,6 +765,16 @@ export class SNApplication {
...
       password
     );
+    /**
+     * Payloads that are waitingForKey, must be imported in order to decrypt them later.
+     */
+    const payloadsWaitingForKeys = decryptedPayloads.filter((payload) => {
+      return payload.waitingForKey && payload.errorDecrypting;
+    });
+    let affectedUuids: string[] = [];
+    if (payloadsWaitingForKeys) {
```
mobitar, line 775: This would always be true. Should check for length.

lib/services/protocol_service.ts
johnny243: Done 👍

lib/application.ts
```diff
@@ -783,16 +793,17 @@ export class SNApplication {
       } else {
         return payload;
       }
-    })
-    const affectedUuids = await this.modelManager!.importPayloads(validPayloads);
+    });
+    const validPayloadsUuids = await this.modelManager!.importPayloads(validPayloads);
```
mobitar, line 797: If you're eventually going to import both valid and invalid payloads, why even separate them in the first place?

johnny243: Because the order matters. If we import everything at once, some payloads may not be decrypted. This makes sure all payloads are imported correctly. 

Additionally:

https://github.com/standardnotes/snjs/blob/a27979314775f7aa34d4817bee740ab3fa3948c5/lib/application.ts#L769

won't have payloads that are `waitingForKeys`. If `waitingForKeys` then `errorDecrypting` is `true` according to [this](https://github.com/standardnotes/snjs/blob/a27979314775f7aa34d4817bee740ab3fa3948c5/lib/services/protocol_service.ts#L527) line.

Another solution would be to import _validPayloads_ first and then _payloadsWaitingForKeys_. But in the end, we can't import them both at the same time.

mobitar: Why would order matter? [This line](https://github.com/standardnotes/snjs/blob/004/lib/services/protocol_service.ts#L146) would monitor items key insertion/changes and decrypt any errorred items. So even if after calling importData not all items are immediately decrypted, they would _eventually_ be decrypted in another run cycle. But it would seem this problem here anyway would not be solved by importing data separately. Instead, it seems like you would want to modify `protocolService.payloadsByDecryptingBackupFile` to decrypt items keys first, then other item types.

lib/services/protocol_service.ts
```diff
@@ -522,6 +522,13 @@ export class SNProtocolService extends PureService implements EncryptionDelegate
     if (format === PayloadFormat.DecryptedBareObject) {
       return payload;
     }
+    /** We check if the key can decrypt the encrypted payload. */
+    if (
+      payload.usesItemKeyEncryption() &&
+      key?.content_type === ContentType.RootKey
+    ) {
+      key = undefined;
```
mobitar, line 530: I think we should throw an exception here. We shouldn't be making up for a caller's misuse of our function. If you pass a payload and the wrong key type, this is a logical error and an exception should be thrown. So we should solve this one level up in the stack trace.

PR #39, opened 2020-07-07
feature: mobile biometrics keychain migration

radko93:


lib/migrations/2020-01-15.ts
```diff
@@ -333,6 +333,9 @@ export class Migration20200115 extends Migration {
       };
       const timing = keychainValue.offline.timing;
       rawStructure.unwrapped![StorageKey.MobilePasscodeTiming] = timing;
+      const biometricPrefs = keychainValue.biometrics_prefs;
+      rawStructure.unwrapped![StorageKey.BiometricPrefs] = { enabled: biometricPrefs.enabled };
```
mobitar, line 337: In the legacy code, biometricPrefs contains both `enabled` and `timing`. Here you are breaking them up into two separate keys. Is this intentional? If so, why? And if to better separate things, then why is BiometricPrefs still an object and not just a boolean?

radko93: If I would keep it one object like before then before saving a value I would need to read the current value every time. So maybe I will just make it a boolean instead of an object then.

radko93: Additionally, if this would an object snjs would need to set a value and be aware of that value which I guess should not be the case?

radko93: @mobitar I added a new commit that changes naming here, `BiometricsPrefs` is now `BiometricsState` and just a boolean.

Issue #20, opened 2020-06-15
Add duplicate_of support


mobitar:
The new syncing server uses a top-level `duplicate_of` field that is equal the to the UUID of the item this item is a duplicate of. Currently production snjs implements this via `item.content.conflict_of`, which is encrypted to the server. In the new behavior, we want to move this value up to the main level (`item.duplicate_of`).

The server uses this to determine how to handle revision history for items that have been conflicted.

cc @karolsojko

karolsojko:
Relates to: https://github.com/standardnotes/syncing-server/pull/60

baptiste-grob:
Let me make sure I understand this correctly:

`item.content.conflict_of` should be used instead of `item.duplicate_of`.

1. What's the compatibility story/what are things I should watch out for?
1. Should snjs handle both fields at the same time, i-e can `item.duplicate_of` be compared with `item.payload.conflict_of`?
1. Should we migrate existing items to this new format, and if so when? Thinking about server load here, but maybe there are other considerations. I'm assuming this migration can only be done on the client since `conflict_of` is part of `item.payload`, which I believe is encrypted (correct me if I'm wrong).

mobitar:
> item.content.conflict_of should be used instead of item.duplicate_of.

Nope, you have it backwards. I've updated the main text to clarify this.

> What's the compatibility story/what are things I should watch out for?

Currently item.content.conflict_of is cleared whenever the item is selected. This way it removes the "Conflicted Copy" flag in the UI. What we'll do is keep item.content.conflict_of as a UI flag. But item.duplicate_of should never be cleared and is permanent. It should be set when an item is duplicated. And should not be used in the UI. It's only a server flag. Anything the UI needs should be in the content.conflict_of field.

> Should snjs handle both fields at the same time, i-e can item.duplicate_of be compared with item.payload.conflict_of?

I think the above should also answer this. We will keep both. One is server-centric, the other is UI centric.

> Should we migrate existing items to this new format, and if so when? Thinking about server load here, but maybe there are other considerations. I'm assuming this migration can only be done on the client since conflict_of is part of item.payload, which I believe is encrypted (correct me if I'm wrong).

We don't need to migrate existing items. This feature is largely related to the new note history feature, and there won't be any history for pre-existing items.

PR #40, opened 2020-07-08
feat: add duplicate_of support

johnny243:
For #20


lib/protocol/payloads/functions.ts
```diff
@@ -65,6 +65,7 @@ export async function PayloadsByDuplicating(
       conflict_of: payload.uuid,
     };
   }
+  override.duplicate_of = payload.uuid;
```
mobitar, line 68: This should be defined under line 60.

lib/protocol/payloads/pure_payload.ts
```diff
@@ -119,6 +120,8 @@ export class PurePayload {
       this.version = (this.content as PayloadContent).version;
     }
 
+    this.duplicate_of = rawPayload.duplicate_of;
```
mobitar, line 123: Should be under line 100.

mobitar:
I forgot to mention that we also want to send this field to the server, save it to disk payloads, and include in backup files the user downloads:

- https://github.com/standardnotes/snjs/blob/5df6851ea4a59b17503ab99991d3d71b3c5d6cf8/lib/protocol/payloads/generator.ts#L121
- https://github.com/standardnotes/snjs/blob/5df6851ea4a59b17503ab99991d3d71b3c5d6cf8/lib/protocol/payloads/generator.ts#L93
- https://github.com/standardnotes/snjs/blob/5df6851ea4a59b17503ab99991d3d71b3c5d6cf8/lib/protocol/payloads/generator.ts#L104

Do we have tests that ensures certain fields are included in certain ejectable situations? For example, if syncing, items_key_id and duplicate_of (amongst others) should be in [ejected](https://github.com/standardnotes/snjs/blob/5df6851ea4a59b17503ab99991d3d71b3c5d6cf8/lib/protocol/payloads/pure_payload.ts#L133) payload.

johnny243:
> I forgot to mention that we also want to send this field to the server, save it to disk payloads, and include in backup files the user downloads:

Done :+1: 

> Do we have tests that ensures certain fields are included in certain ejectable situations? For example, if syncing, items_key_id and duplicate_of (amongst others) should be in [ejected](https://github.com/standardnotes/snjs/blob/5df6851ea4a59b17503ab99991d3d71b3c5d6cf8/lib/protocol/payloads/pure_payload.ts#L133) payload.

I found [this](https://github.com/standardnotes/snjs/blob/4a8577bc148fbcbd0d775d9a55606b54f71dbec1/test/protocol.test.js#L79) test, which is just making sure there are no meta-data like `fields`, `source` and `format` when the payload is `ejected()`.

In order to write tests for this situations, should I use the `payloadsByEncryptingPayload` method from `protocolService` and pass the corresponding [EncryptionIntent](https://github.com/standardnotes/snjs/blob/004/lib/protocol/intents.ts) which represents the situation?

Example:

```javascript
const payload = { /* Generate payload */ };
const encryptedPayload = await this.application.protocolService.payloadByEncryptingPayload(
  payload,
  EncryptionIntent.Sync // For the syncing situation
);
const ejected = encryptedPayload.ejected();
expect(ejected).to.be.ok
expect(ejected).to.have.keys('items_key_id', 'duplicate_of', /* ... and other expected fields to be ejected */);
```

mobitar:
In this case you would be testing the function `payloadByEncryptingPayload` to ensure it produces the correct result. Which perhaps should be a test of its own. But for now you should specifically target the underlying function that `payloadByEncryptingPayload` uses to create payloads.

PR #41, opened 2020-07-09
Better protocol upgrade api + blocking dialog

baptiste-grob:
- Makes SNAlertService a type which clients have to pass an implementation of in the constructor
- Adds `blockingDialog` method to SNAlertService
- Lets clients clearly know if the protocol upgrade succeeded, failed, or canceled.
- Shows a blocking dialog during the encryption upgrade process (likely needs a better message)

### DIFF
--- a/lib/application.ts
+++ b/lib/application.ts
@@ -683,10 +683,11 @@ export class SNApplication {
     return !isNullOrUndefined(this.getUser()) || this.hasPasscode();
   }
 
-  /**
-   * @returns An array of errors, if any.
-   */
-  public async upgradeProtocolVersion() {
+  public async upgradeProtocolVersion(): Promise<{
+    success?: true,
+    canceled?: true,
+    error?: Error,
+  }> {
     const hasPasscode = this.hasPasscode();
     const hasAccount = !this.noAccount();
     const types = [];
@@ -699,7 +700,7 @@ export class SNApplication {
     const challenge = new Challenge(types, ChallengeReason.ProtocolUpgrade);
     const response = await this.challengeService!.promptForChallengeResponse(challenge);
     if (!response) {
-      return;
+      return { canceled: true };
     }
     let passcode: string | undefined;
     if (hasPasscode) {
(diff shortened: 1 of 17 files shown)


lib/application.ts
```diff
@@ -683,10 +689,11 @@ export class SNApplication {
     return !isNullOrUndefined(this.getUser()) || this.hasPasscode();
   }
 
-  /**
-   * @returns An array of errors, if any.
-   */
-  public async upgradeProtocolVersion() {
+  public async upgradeProtocolVersion(): Promise<{
+    success?: true,
```
mobitar, line 693: Wait why does this default to true?

lib/services/api/messages.ts
```diff
@@ -45,6 +45,7 @@ export const INVALID_PASSWORD                      = `Invalid password.`;
 
 export const OUTDATED_PROTOCOL_ALERT_TITLE         = 'Update Recommended';
 export const OUTDATED_PROTOCOL_ALERT_IGNORE        = 'Sign In';
+export const UPGRADING_ENCRYPTION                  = 'Upgrading the account encryption. Do not close the application.';
```
mobitar, line 48: "Your account's encryption version is being upgraded. Do not close the application until this process completes."

lib/application.ts
baptiste-grob: this isn't a default, it's the `true` type. So this says that if `success` is present, it's going to be `true`. an alternative would have been to have the success property be mandatory, and a boolean. But it just resulted in noise wherever we would `return { ... }`.

baptiste-grob:
Heads up @radko93, this is a breaking change, you'll have to update your SNApplication constructor call and add/change a couple of things in your AlertService. Feel free to reach out if anything is left unclear after looking at the new types :)

lib/application.ts
mobitar: Ah interesting. So TS would complain if you tried to return `{success: false}`?

baptiste-grob: Yes, exactly!

Issue #13, opened 2020-05-20
Keychain interface should be namespaced


mobitar:
As part of device interface/protocol service. Currently it is not equipped to handle multiple applications instances writing to the keychain. The last application to write replaces any previous values.

PR #42, opened 2020-07-11
feat: namespaced keychain

johnny243:
For #13


lib/device_interface.ts
```diff
@@ -29,6 +30,7 @@ export abstract class DeviceInterface {
     this.namespace = namespace;
     this.timeout = timeout || setTimeout.bind(getGlobalScope());
     this.interval = interval || setInterval.bind(getGlobalScope());
+    this.keychainStorageKey = this.keyForKeychainStorage();
```
mobitar, line 33: Doesn't need to be cached. Function can be called on demand.

PR #43, opened 2020-07-13
fix: uninstall active extensions

baptiste-grob:
The component manager's `deregisterComponent` method calls sync. Because sync is a very high-level operation with a lot of side-effects it can conflict when we're trying to do something after calling deregisterComponent. In this case, uninstalling it, which will be invalidated by the syncing request completing and determining that the just-disabled component should be kept.
I also think that in general calling sync inside a method named like this is surprising behavior. At least I expected it to do much less work when looking at it.


baptiste-grob:
Thanks, I'm keeping this in mind.

Issue #38, opened 2020-07-06
Server history support


mobitar:
Consumes this API: https://github.com/standardnotes/syncing-server/pull/60

Session history and Server history should be combined into a single HistoryService interface. Clients should implement a single History menu that separates into two sections: Server History, and Session History. Selecting a revision from either section should present the same UI.

cc @radko93 @baptiste-grob @johnny243

johnny243:
@mobitar when you say Session history, are you referring to [user sessions](https://github.com/standardnotes/syncing-server/blob/develop/app/controllers/api/sessions_controller.rb)?

mobitar:
No, this session history:

<img width="391" alt="Screen Shot 2020-07-08 at 8 42 58 AM" src="https://user-images.githubusercontent.com/3277844/86925912-0d406300-c0f7-11ea-8578-2d85a22b9a17.png">

PR #44, opened 2020-07-14
feat: server history support

johnny243:
For #38


lib/services/history/history_server.ts
```diff
@@ -0,0 +1,31 @@
...
+      uuids.forEach((itemUUID) => {
+        const rawItemHistory = content.itemUUIDToItemHistoryMapping[itemUUID];
+        content.itemUUIDToItemHistoryMapping[itemUUID] =
+          ItemHistory.FromJson(rawItemHistory);
+      });
+      return new HistoryServer(content);
+    } else {
+      return new HistoryServer();
+    }
+  }
```
johnny243, line 30: TODO: this method should create an instance of `HistoryServer` from the item's revision response.

- [ ] It should parse the response object
- [ ] It should decrypt payloads

johnny243: cc @mobitar

mobitar: The naming HistoryServer implies to me this is a server of some sort. If that's not the case, this should be called 'ServerHistoryItem' or something like that.

mobitar: Or ServerHistoryRevision.

lib/services/history/history_manager.ts
```diff
@@ -141,10 +163,20 @@ export class SNHistoryManager extends PureService {
...
   }
 
-  historyForItem(item: SNItem) {
+  sessionHistoryForItem(item: SNItem) {
     return this.historySession!.historyForItem(item.uuid);
   }
 
+  async fetchHistoryFromServer(item: SNItem) {
+    const itemRevisionsResponse = await this.apiService!.getItemRevisions(item.uuid);
+    this.historyServer = await ServerHistoryRevision.FromResponse(this.protocolService!, item.uuid, itemRevisionsResponse);
```
johnny243, line 172: I had to pass the item's `uuid` in order to decrypt revisions. In my opinion, the server response should be the following:

- `uuid`: the `uuid` of the item
- `revision_uuid`: the `uuid` of the revision

lib/services/api/api_service.ts
```diff
@@ -314,4 +316,41 @@ export class SNApiService extends PureService {
...
+    ).catch((errorResponse: HttpResponse) => {
+      return this.errorResponseWithFallbackMessage(
+        errorResponse,
+        messages.API_MESSAGE_GENERIC_SYNC_FAIL
+      );
+    });
+    return response;
+  }
+
+  async getRevisionForItem(itemId: string, revisionId: string) {
```
johnny243, line 338: Since all revisions are returned with `getItemRevisions`, I didn't require to use this endpoint in the end.

lib/protocol/payloads/generator.ts
```diff
@@ -167,6 +167,8 @@ const ServerSavedPayloadFields = [
   PayloadField.LastSyncEnd
 ]
 
+const ServerHistoryPayloadFields = ServerPayloadFields
```
mobitar, line 170: I would do `ServerPayloadFields.slice()` just to be safe.

lib/services/api/api_service.ts
```diff
@@ -314,4 +316,41 @@ export class SNApiService extends PureService {
     return result;
   }
 
+  async getItemRevisions(itemId: string) {
+    if (!this.session) {
+      return undefined;
```
mobitar, line 321: Shouldn't we return an error response, like we do in sync and all other API requests? Also, what if `this.refreshingSession` is true here? Maybe we should make a general function for basic checks, like this:

```javascript
preprocessingError() {
  if (this.refreshingSession) {
    return this.createErrorResponse(messages.API_MESSAGE_TOKEN_REFRESH_IN_PROGRESS);
  }
  if (!this.session) {... }

  return undefined;
}
```

Then in all API requests, you do:

```javascript
const preprocessingError = this.preprocessingError();
if (preprocessingError) { return preprocessingError }
```

lib/services/api/api_service.ts
mobitar: In this case we should modify the server to achieve the desired result. If an item with a 1MB content field has 400 revisions, `getItemRevisions` would be a 400MB request to the server, and it would probably crash the client.

We should modify /revisions endpoint on the server to only return metadata about the revision, and then use /revision/:id to download each revision when the user actually selects it.

lib/services/history/base_history.ts
```diff
@@ -0,0 +1,34 @@
+import { PurePayload } from '@Payloads/pure_payload';
+import { ItemHistory } from '@Services/history/item_history';
+
+export type HistoryContent = {
+  itemUUIDToItemHistoryMapping: Record<string, ItemHistory>
```
mobitar, line 5: I think you over-extracted some things from SessionHistory.

Session history was meant to be an in-memory data store. So we had to use data structures to map uuids to item histories. This is not functionality we would want in an abstract class however. I'm actually not sure this abstract class is required at all tbh. If there are common functions we want, we should just define an `interface` instead.

lib/services/history/history_manager.ts
```diff
@@ -5,26 +5,33 @@ import { SNItem } from '@Models/core/item';
...
 export class SNHistoryManager extends PureService {
 
   private itemManager?: ItemManager
   private storageService?: SNStorageService
+  private apiService?: SNApiService
+  private protocolService?: SNProtocolService
   private contentTypes: ContentType[] = []
   private timeout: any
   private historySession?: HistorySession
+  private historyServer?: ServerHistoryRevision
```
mobitar, line 34: The naming for historySession makes sense because it is quite literally a session—a transient structure holding data. But with `historyServer`, that naming doesn't make sense. This should be named something else. Also, I don't even see why you need a property here? When a client accesses server history, there should be no state saved—it's all a matter of API requests, and we simply relay results to the client. The client then displays those results. When the client dismisses the history menu, the history data vanishes.

lib/services/history/history_manager.ts
```diff
@@ -120,9 +133,18 @@ export class SNHistoryManager extends PureService {
     this.historySession!.setItemRevisionThreshold(threshold);
   }
 
-  async addHistoryEntryForItem(item: SNItem) {
-    const payload = CreateSourcedPayloadFromObject(item, PayloadSource.SessionHistory)
-    const entry = this.historySession!.addEntryForPayload(payload);
+  addEntryForPayload(payload: PurePayload) {
+    switch (payload.source) {
+      case PayloadSource.ServerHistory:
+        return this.historyServer!.addEntryForPayload(payload);
```
mobitar, line 139: I don't think this is necessary. We should just return to the client an array of revisions. No need for this encapsulation. The history server objet should not exist.

lib/services/history/server_history_revision.ts
```diff
@@ -0,0 +1,35 @@
...
+      Object.entries(response).forEach(([key, value]) => revisions.push(value));
+      revisions.sort((a, b) => {
+        return a.updated_at! < b.updated_at! ? -1 : 1;
+      }).map(async (revision) => {
+        const payload = CreateSourcedPayloadFromObject(revision, PayloadSource.ServerHistory, {
+          ...revision,
+          uuid: itemUuid
+        });
+        const decryptedPayload = await protocolService.payloadByDecryptingPayload(payload);
+        historyServer.addEntryForPayload(decryptedPayload);
```
mobitar, line 28: See prior comments. This shouldn't exist.

lib/services/history/history_manager.ts
```diff
@@ -141,10 +163,20 @@ export class SNHistoryManager extends PureService {
     }
   }
 
-  historyForItem(item: SNItem) {
+  sessionHistoryForItem(item: SNItem) {
     return this.historySession!.historyForItem(item.uuid);
   }
 
+  async fetchHistoryFromServer(item: SNItem) {
```
mobitar, line 170: `fetchHistoryFromServer` and `serverHistoryForItem` should be combined into one function called `serverHistoryForItem` which returns an array of revisions.

lib/services/history/item_history.ts
```diff
@@ -19,17 +19,19 @@ export class ItemHistory {
   constructor(entries?: ItemHistoryEntry[]) {
     /** Deserialize the entries into entry objects. */
     if (entries) {
-      for (const entry of entries) {
+      entries.slice(0).sort((a, b) => {
```
mobitar, line 22: What does slice(0) do? I could google it, but that's the point. Should be more clear.

lib/services/history/server_history_revision.ts
```diff
@@ -0,0 +1,35 @@
...
+import { PayloadSource } from '@Lib/protocol/payloads';
+import { SNProtocolService } from '../protocol_service';
+
+export class ServerHistoryRevision extends BaseHistory {
+
+  constructor(content?: HistoryContent) {
+    super(content);
+  }
+
+  static async FromResponse(protocolService: SNProtocolService, itemUuid: string, response?: HttpResponse) {
```
mobitar, line 13: Passing a service to a low level object is not good style. Objects should be self contained and should be fed information from a service. This needs to be extracted one level up.

lib/services/history/history_manager.ts
mobitar: Why would we overwrite `this.historyServer` with item-specific data every time we fetch a new item's revision? This is overly extracted from the session history design. These two are not the same and should not be treated as such.

test/history.test.js
```diff
@@ -20,210 +20,334 @@ describe('session history', () => {
...
+      let itemHistory = await this.historyManager.serverHistoryForItem(item);
+
+      /** Server history should not save initial revision */
+      expect(itemHistory).to.be.ok;
+      expect(itemHistory.entries.length).to.equal(0);
+
+      /** Sync with same contents, should not create new entry */
+      await this.application.saveItem(item.uuid);
+      itemHistory = await this.historyManager.serverHistoryForItem(item);
+      expect(itemHistory.entries.length).to.equal(0);
```
johnny243, line 277: This assertion is failing. The server returns a revision if the item is saved (with no changes) and synced. Ideally we should save space by not saving a revision if there are no changes.

lib/services/history/item_history_entry.ts
```diff
@@ -14,8 +18,9 @@ export class ItemHistoryEntry {
   protected defaultContentKeyToDiffOn = 'text'
   protected textCharDiffLength = 0
   protected hasPreviousEntry = false
+  protected source: ItemHistorySource
```
mobitar, line 21: Should be marked as `readonly` as well if we don't plan on changing this value after the constructor.

lib/services/history/history_manager.ts
```diff
@@ -190,4 +198,12 @@ export class SNHistoryManager extends PureService {
       );
     }
   }
+
+  async serverHistoryForItem(item: SNItem) {
```
mobitar, line 202: Should be named `remoteHistoryForItem` for consistency.

lib/services/history/item_history_entry.ts
```diff
@@ -39,7 +45,7 @@ export class ItemHistoryEntry {
     /** We'll try to compute the delta based on an assumed
      * content property of `text`, if it exists.
      */
-    if (this.payload.contentObject[this.defaultContentKeyToDiffOn]) {
+    if (this.payload.content && this.payload.contentObject[this.defaultContentKeyToDiffOn]) {
```
mobitar, line 48: You can use `this.payload.safeContent` instead of doing your own safety check.

lib/services/history/item_history_entry.ts
```diff
@@ -100,4 +106,8 @@ export class ItemHistoryEntry {
     const rhs = CreateItemFromPayload(entry.payload);
     return lhs.isItemContentEqualWith(rhs);
   }
+
+  isFromServer() {
```
mobitar, line 110: Should be `isRemoteSource` for consistency.

lib/services/history/sources/remote_history.ts
```diff
@@ -0,0 +1,71 @@
...
+  private protocolService: SNProtocolService
+
+  constructor(apiService: SNApiService, protocolService: SNProtocolService) {
+    this.apiService = apiService;
+    this.protocolService = protocolService;
+  }
+
+  /**
+   * Iterates over the response and creates a payload from each entry.
+   * @param response
```
mobitar, line 20: `@param` not necessary with TypeScript unless you're adding a description.

lib/services/history/sources/remote_history.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  private responseToPayloadArray(response: HttpResponse) {
+    let revisions: RawPayload[] = response!.revisions;
```
mobitar, line 23: Should be `const`

lib/services/history/sources/remote_history.ts
```diff
@@ -0,0 +1,71 @@
...
+      };
+    }
+    const payloadArray = this.responseToPayloadArray(serverResponse);
+    const itemHistory = ItemHistory.FromJson(payloadArray, ItemHistorySource.Remote);
+    return itemHistory;
+  }
+
+  /**
+   * Creates a PurePayload from the response.
+   * @param response
```
mobitar, line 48: Not necessary

lib/services/history/sources/remote_history.ts
```diff
@@ -0,0 +1,71 @@
...
+   */
+  private responseToPayload(response: HttpResponse) {
+    delete response.error;
+    delete response.status;
```
mobitar, line 52: Should definitely not be mutating input response object, especially unexpectedly because you don't know the caller doesn't require these properties.

These properties should be ignored by `CreateMaxPayloadFromAnyObject` anyway since they're not a `PayloadField`.

You can also try using something like `response as Partial<RawPayload>` instead of `unknown`.

test/history.test.js
mobitar: Session history has access to decrypted objects, so can do equality comparison to prevent redundant revisions. However server doesn't have access to decrypted content, so can't do equality comparison. So this is expected, and this should be updated to reflect a length of 1.

baptiste-grob:
Removing myself from review as I'm not familiar enough with the existing code to provide any value here

lib/services/history/sources/remote_history.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+    const payloadArray = this.responseToPayloadArray(serverResponse);
+    const itemHistory = ItemHistory.FromJson(payloadArray, ItemHistorySource.Remote);
+    return itemHistory;
+  }
+
+  /**
+   * Creates a PurePayload from the response.
+   */
+  private responseToPayload(response: RevisionResponse) {
```
mobitar, line 60: This function appears to only be used once, and is short enough to just be inline.

lib/services/history/sources/remote_history.ts
```diff
@@ -0,0 +1,78 @@
...
+
+  constructor(apiService: SNApiService, protocolService: SNProtocolService) {
+    this.apiService = apiService;
+    this.protocolService = protocolService;
+  }
+
+  /**
+   * Iterates over the response and creates a payload from each entry.
+   */
+  private responseToPayloadArray(response: RevisionsResponse) {
```
mobitar, line 33: This should also be inline, so we don't lose context.

lib/services/history/sources/remote_history.ts
```diff
@@ -0,0 +1,78 @@
...
+    const revisions = response.revisions!;
+    const revisionEntries = revisions.map((revision) => {
+      return {
+        payload: CreateMaxPayloadFromAnyObject(revision)
```
mobitar, line 37: Wouldn't the payload.uuid at this point be the uuid of the revision and not the item? We don't seem to be replacing the uuid like we do in `fetchItemRevision`.

If this is the case, you should:

1. Create a test that catches this issue.
2. Create a reusable function `rawRevisionToItemPayload` that cleans/restructures revision responses as needed and creates payloads from them.

lib/services/history/sources/remote_history.ts
```diff
@@ -0,0 +1,78 @@
...
+import { SNProtocolService } from "@Lib/services/protocol_service";
+import { ItemHistory } from "../item_history";
+import { HttpResponse } from "@Lib/services/api/http_service";
+import { RawPayload, CreateMaxPayloadFromAnyObject, CreateSourcedPayloadFromObject } from "@Lib/protocol/payloads/generator";
+import { ItemHistorySource, ItemHistoryEntry } from "../item_history_entry";
+import { PayloadSource } from "@Lib/protocol/payloads";
+
+type RevisionResponse = {
+  error?: any
+  revision?: RawPayload
```
mobitar, line 11: Revisions are not exactly RawPayloads. They share 99% of the same properties, but they still differ in that a revision has an `item_uuid` property. You should instead create a new RawRevision type that extends RawPayload.

lib/services/history/sources/remote_history.ts
johnny243: > Wouldn't the payload.uuid at this point be the uuid of the revision and not the item? We don't seem to be replacing the uuid like we do in fetchItemRevision.

Yes, but we don't need to replace it in this case, the revision doesn't have a `content` field and thus will not be decryptable.

mobitar: Ok, then we need to create a type that describes exactly what the response is, to avoid this kind of confusion. It should be very strictly defined what a `revisions` element looks like.

johnny243: This should be the types:

```javascript
type RawRevision = RawPayload & {
  item_id: string
}
```

```javascript
type RawRevisionsListItem = {
  uuid: string
  content_type: ContentType
  created_at: string
  updated_at: string
}
```

`getItemRevisions` should return an array of `RawRevisionsListItem`.

But since the body fields are combined with the top level response object, I will have to parse the response and iterate the object keys.

PR #45, opened 2020-07-14
Defensive component mananger

baptiste-grob:
When finding components by UUIDs in the component manager, do not assume a component will always be there. This can happen when a client queries or wants to update a component that's been deleted by another client, for example.
Instead, rely on the component's UUID as much as possible when making changes.


baptiste-grob:
Just noticed this causes a regression with some cases (tags). Turning this into a draft.

mobitar:
Ready for review?

baptiste-grob:
Yes!

lib/services/component_manager.ts
```diff
@@ -112,6 +112,30 @@ type ComponentState = {
   sessionKey?: string
 }
 
+class ActiveComponents {
```
mobitar, line 115: Some thoughts on this class:

- The properties should be private and readonly. You should instead expose functions for querying. So instead of `this.activeComponents.uuids.includes` you would do `this.activeComponents.has(uuid)`.
- I don't think you need two separate data structures, array and record. You just need record. There seems to be an overlap here. The `add` function would just remove the if conditional. The `remove` function would remove line 128. And the `clear` function would remove line 135. Then you'd introduce a function called `has` that checks to see if `this.areas[uuid] !== undefined`.
- Is this class really required? Sounds like you could just do the same thing with a Record property. Or maybe a Set property. It does offer some abstraction I suppose.

lib/services/component_manager.ts
```diff
@@ -183,7 +207,7 @@ export class SNComponentManager extends PureService {
     super.deinit();
     this.streamObservers.length = 0;
     this.contextStreamObservers.length = 0;
-    this.activeComponents.length = 0;
+    this.activeComponents.clear();
```
mobitar, line 210: `activeComponents ` should just be nulled out. We don't need a recycle method. The story is that `deinit` used to be a method for recycling a service, but then we moved towards complete nuking of a service. So you could do away with the clear function altogether and just set activeComponents to undefined.

lib/services/component_manager.ts
baptiste-grob: I moved away from a Record because I had made an assumption that I realize is no longer true anymore. I'll walk that back.

PR #47, opened 2020-07-15
feature: expose isMobileDefault to ComponentMutator

radko93:


radko93:
@baptiste-grob I don't run snjs locally always when building mobile so it's better for me to update dist files

PR #49, opened 2020-07-16
Feature/conflict resolution while editing

baptiste-grob:
When two items have different contents:

- If the item was edited in the last 20 seconds, keep it, and duplicate the incoming item
- If the item was edited more than 20 seconds ago, duplicate it and keep the incoming item

Should we do this when an incoming item has been deleted as well?

### DIFF
--- a/lib/models/core/item.ts
+++ b/lib/models/core/item.ts
@@ -284,15 +284,23 @@ export class SNItem {
     if (!contentDiffers) {
       return ConflictStrategy.KeepRight;
     }
-    const differsExclRefs = ItemContentsDiffer(
+    const itemsAreDifferentExcludingRefs = ItemContentsDiffer(
       this,
       item,
       ['references']
     );
-    if (differsExclRefs) {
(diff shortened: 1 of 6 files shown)


lib/models/core/item.ts
```diff
@@ -284,15 +284,23 @@ export class SNItem {
...
+    if (itemsAreDifferentExcludingRefs) {
+      const twentySeconds = 20_000;
+      if (Date.now() - this.userModifiedDate.getTime() < twentySeconds ) {
+        /**
+         * The user may be editing our item at this time, so duplicate the incoming
+         * item to avoid creating surprises in the client's UI.
+         */
+        return ConflictStrategy.KeepLeftDuplicateRight;
+      }
+      return ConflictStrategy.DuplicateLeftKeepRight;
```
mobitar, line 301: I like an `else` here even though it's redundant. It makes it clear that there are only two possible conditions. What do you think?

baptiste-grob: I was indecisive. Thanks for weighing in :)

PR #52, opened 2020-07-22
fix: getJsonParsedStorageValue

johnny243:
To prevent passing an `object` to `JSON.parse` which throws the below error:

```
Uncaught SyntaxError: Unexpected token o in JSON at position 1
    at JSON.parse (<anonymous>)
```


mobitar:
In what cases would an object exist in raw storage? Does AsyncStorage in React Native allow saving objects directly? @radko93

johnny243:
Relevant line: https://github.com/standardnotes/mobile/blob/d59b5dd538108054fdfc3ef560dfed154d32b179/src/lib/interface.ts#L74

mobitar:
I think the issue is that the mobile implementation of getRawStorageValue should not be JSON parsing. That's the entire purpose of `raw`. So this should indeed fail and we do not want to catch this exception. We should modify the interface implementation of raw to not JSON parse.

mobitar:
@radko93 SNJS expects these raw interface functions to return raw strings. I suppose we should make that more explicit in the declaration since that's what raw could only mean. Or it could be a number or boolean I suppose. So maybe the return value is `string | boolean | number | undefined`.

radko93:
AsyncStorage expects a string https://react-native-community.github.io/async-storage/docs/api#setitem

radko93:
@johnny243 can you add changes to your PR?

johnny243:
Closing this. We should fix mobile implementation instead: https://github.com/standardnotes/mobile/pull/273

mobitar:
Maybe https://github.com/standardnotes/mobile/pull/273 is not necessary actually. I think the behavior is fine, since the set/get handles JSON internally as an implementation detail. In that case I'm not quite sure why an object is being returned here. In the new code, consumers of setRaw would pass in a JSON string, mobile would JSON stringify again, and save the value (so it's double stringified).  On get, it would JSON parse once, and return the result. This means it still needs one more layer of parsing. So it wouldn't cause any exceptions in the code in this PR.

However with legacy values, it's possible it wasn't double stringified. Thus this exception. So you can keep this code as-is, but add a comment explaining why this would happen.

Also I believe the code can be simplified to:

```javascript
    try {
      return JSON.parse(value);
    } finally {
      return value;
    }
```

baptiste-grob:
A `finally` block would short-circuit the actual return value but you could definitely put `return value;` inside the `catch` block since we're only expecting it to happen in this instance

mobitar:
> would short-circuit the actual return value

You mean it would override the return inside try {}?

baptiste-grob:
> You mean it would override the return inside try {}?

Yes.

mobitar:
Hmm that's strange. I wouldn't expect that. But you're right:

```javascript
const fn = () => {
    try {
        return 1;
    } finally {
        return 2;
    }
}

const value = fn();
console.log(value);
```

This prints `2`. But I would have thought the above code would essentially be equivalent to:

```javascript
return 1;
return 2;
```

mobitar:
Are we scrapping this PR?

johnny243:
I believe we should keep it:

> However with legacy values, it's possible it wasn't double stringified. Thus this exception. So you can keep this code as-is, but add a comment explaining why this would happen.

PR #54, opened 2020-07-28
feature: add support for a "not" operator

rsesek:
A not operator would allow you to build more powerful filters, which would be especially useful when dealing with an imported a tag hierarchy from another system that you're trying to clean up.

This was [discussed here](https://github.com/standardnotes/forum/issues/262#issuecomment-490623566). But rather than introducing a special syntax to parse, this just treats it as a top-level operator like and/or.

I couldn't find [the docs](https://standardnotes.org/help/42/how-do-i-view-a-list-of-untagged-notes-and-create-other-dynamic-filters) to update on Github. That should be done if this PR is accepted too.

### DIFF
--- a/lib/models/core/predicate.ts
+++ b/lib/models/core/predicate.ts
@@ -105,6 +105,11 @@ export class SNPredicate {
       targetValue = this.DateFromString(targetValue);
     }
 
+    /* Process not before handling the keypath, because not does not use it. */
+    if (predicate.operator === 'not') {
+      return !this.ObjectSatisfiesPredicate(object, targetValue);
+    }
+
     const valueAtKeyPath = predicate.keypath.split('.').reduce((previous, current) => {
       return previous && previous[current];
     }, object);
--- a/test/predicate.test.js
+++ b/test/predicate.test.js
@@ -120,6 +120,51 @@ describe('predicates', async function () {
     ]))).to.equal(false);
   });
 
+  it('test not operator', async function () {
+    const item = await this.createItem();
+    expect(item.satisfiesPredicate(new SNPredicate(
+      'this_field_ignored', 'not', ['content.title', '=', 'Not This Title']
+    ))).to.equal(true);
+    expect(item.satisfiesPredicate(new SNPredicate(
+      'this_field_ignored', 'not', ['content.title', '=', 'Hello']
+    ))).to.equal(false);
+
+    expect(item.satisfiesPredicate(new SNPredicate('', 'and', [
+      ['', 'not', ['content.tags', 'includes', ['title', '=', 'far']]],
+      ['content.tags', 'includes', ['title', '=', 'foo']],
+    ]))).to.equal(false);
+    expect(item.satisfiesPredicate(new SNPredicate('', 'and', [
+      ['', 'not', ['content.tags', 'includes', ['title', '=', 'boo']]],
+      ['content.tags', 'includes', ['title', '=', 'foo']],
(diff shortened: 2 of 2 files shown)


mobitar:
Hmm, why would `not` be a recursive operator and not just a standard operator that falls in the same category as `=`, `>`, `<`, etc? That is, the operator would be `!=`. You could then use the recursive operators `and`/`or` for predicates that individually use `!=`.

mobitar:
Actually on second look it appears we already have this operator: https://github.com/standardnotes/snjs/blob/a6871434bd533c11d62a64a8330a15022f7c5a4b/lib/models/core/predicate.ts#L139

rsesek:
`not` could be an unary operator rather than a recursive one, and the new patch switches it. I chose the recursive option initially because it does need to recurse.

I don't think `!=` is quite the same because it does not recurse on the value here when there is not a valid key path: https://github.com/standardnotes/snjs/blob/a6871434bd533c11d62a64a8330a15022f7c5a4b/lib/models/core/predicate.ts#L125

That prevents you from using it as a simple negation without a LHS, which I think you need if want to say:

`['', 'not', ['content.tags', 'includes', ['title', '=', 'boo']]]`

E.g., if you replace `not` with `!=` in the new tests, they do not pass.

mobitar:
Ah, I see what you mean. `not` can be applied on an entire predicate, whereas `!=` exists within the predicate. In that case I suppose it could qualify as a recursive operator. But now that you've force pushed I can't recall how it was handled previously 🤔 Which solution do you prefer overall?

rsesek:
The old patch is [here](https://github.com/standardnotes/snjs/commit/a6871434bd533c11d62a64a8330a15022f7c5a4b).

I think the current unary approach is a little better, even if it's not treating `not` as a recursive operator, because creating a single not predicate is a little more intuitive:

`[ '', 'not', ['title', '=', 'Test'] ]`

Whereas when it was a true recursive operator, the value had to be an array of predicates:

`['', 'not', [ ['title', '=', 'Test'] ] ]`

I think the latter might be a little confusing for simple predicates.

(Sorry about the force push; some projects mandate it for a clean history, others don't).

mobitar:
Looks great. Could I also ask you to submit a PR for our docs with info about the availability of this new operator, as well as some examples? https://github.com/standardnotes/docs/blob/master/docs/usage/tags.md

cc @TheodoreChu

rsesek:
Sure! Done at standardnotes/docs#9.

mobitar:
Thanks! We're actively working on finishing up 004 development, so this will ship when that ships 👍

rsesek:
Great, thank you! Looking forward to the update.

mobitar:
@rsesek the PR was reverted as it does not pass type checking: https://github.com/standardnotes/snjs/commit/316b1b0caa00bb2ad5dde07b3a473c1dba2a65c7

`targetValue` is of type `PredicateValue`, but `ObjectSatisfiesPredicate` second parameter expects type `PredicateType`. Please ensure `npm run tsc` passes (or ideally use an editor that supports TypeScript type checking inline).

rsesek:
Apologies for the breakage. Reland with the right types at #59.

It might be good to have a Github action for `tsc` or to run the type checker with `start`.

mobitar:
Yup we have this now :) https://github.com/standardnotes/snjs/commit/a3354a888e299216fa7efcf5a4d36641a7496a1b

PR #55, opened 2020-07-29
feat: namespacing

johnny243:
See: https://github.com/standardnotes/mobile/pull/269#discussion_r461563526

- [x] rename `getKeychainValue` to `getNamespacedKeychainValue`
- [x] rename `setKeychainValue` to `setNamespacedKeychainValue`
- [x] rename `clearKeychainValue` to `clearNamespacedKeychainValue`
- [x] introduce `getRawKeychainValue` and `clearRawKeychainValue` to be used on mobile
- [x] update migration test for web
- [x] update migration test for mobile
- [x] update storage test
- [x] namespace service
- [x] update web_device_interface for tests

`DeviceInterface` for each platform needs to be updated.


test/lib/web_device_interface.js
```diff
@@ -79,16 +79,30 @@ export default class WebDeviceInterface extends DeviceInterface {
     }
   }
 
-  /** @keychian */
-  async getKeychainValue() {
-    return this.keychainValue;
+  /** @keychain */
+  async getNamespacedKeychainValue() {
+    return this.keychainValue[this.keychainStorageKey];
```
mobitar, line 84: So actually I'm thinking that `keychainStorageKey` is no longer required (in general). Instead, this function would just return `return this.keychainValue[this.namespace];` Because the implementation of keychainStorageKey is:

```javascript
  protected get keychainStorageKey() {
    const keychainStorageKey = 'keychain';
    if (this.namespace) {
      return `${this.namespace}-${keychainStorageKey}`;
    } else {
      return keychainStorageKey;
    }
  }
```

But, in what cases would namespace be undefined? It should never be undefined.

I think essentially we got confused with a lot of different concepts. This should be the final structure:

- Each client stores one main keychain value with the storage key `keychain`. For web, this value would be stored in localStorage. For mobile, this value would be stored in the device keychain.
- The `keychain` value is a map: [namespace : value].

mobitar: Also, anytime we call `setNamespacedKeychainValue`, we must also persist the root structure to the device keychain. Because `this.keychainValue` is just an in-memory structure.

johnny243: > But, in what cases would namespace be undefined? It should never be undefined.

The namespace is initialized as an empty string [here](https://github.com/standardnotes/web/blob/79d48b7e077394dcf577438bcc54720ceddb8a87/app/assets/javascripts/ui_models/application.ts#L59).

The below block of code will return 'keychain':

```javascript
protected get keychainStorageKey() {
  const keychainStorageKey = 'keychain';
  if (this.namespace) {
    return `${this.namespace}-${keychainStorageKey}`;
  } else {
    return keychainStorageKey;
  }
}
```

johnny243: But I get the general idea, it looks much better. Maybe we can validate the `namespace` to be required and not be an empty string? Also, we need to update the `web` and `mobile` apps to have a valid namespace.

mobitar: The empty string was essentially a placeholder. We need to dynamically determine an automatic namespace to use, based on how many existing namespaces there are. So, web should pass undefined as namespace, and at that point, SNJS should automatically assign a dynamic namespace that is persisted.

Let me know what you think are good ways to solve this. One idea:

- We could have a storage key `namespaces` that is an array of namespaces: [app-1, app-2, app-3]. Then when creating a new app, we retrieve the last namespace in that array, and increment it to get a new namespace: app-4.

You could build this functionality into this existing PR.

mobitar: We need to also consider the UX of this. If a user has multiple accounts signed into a client, and each account uses its own keychain/storage namespace, how will a user know which account they are switching into if the account data (email address, etc) is encrypted? One solution would be that each namespace stores a user-defined "label" that allows them to distinguish one namespace from the other. This value would not be encrypted and would be stored in root storage in the `namespaces` struct.

johnny243: We can also generate an automatic namespace by taking the length of `namespaces` and increment it. 

But when will SNJS automatically assign a dynamic namespace? When a new account is signed in?

mobitar: If namespaces.length == 0, create new namespace;
If namespaces.length == 1, clients should automatically sign in to that namespace;
If namespaces.length > 1, clients should display UI for selecting which namespace to sign into (this is post-3.5, maybe like 3.6 or 3.7).

So in general it would seem like each namespace value should be an object, so that we can have properties like `label`, `isDefault`, etc.

johnny243: I see. We will put this feature on hold for now, we'll discuss this further once we get back on this.

lib/services/namespace_service.ts
```diff
@@ -0,0 +1,139 @@
...
+    const namespaceIndex = namespaces.findIndex(n => n.identifier === namespace.identifier);
+    if (namespaceIndex === -1) {
+      namespaces.push(namespace);
+    } else {
+      namespaces[namespaceIndex] = namespace;
+    }
+    await this.setNamespaces(namespaces);
+  }
+
+  private async setCurrentNamespace(namespace: SNNamespace) {
```
mobitar, line 58: This should also be called `switchToNamespace`, as I think it's clearer.

lib/services/namespace_service.ts
```diff
@@ -0,0 +1,139 @@
...
+    this.namespace = namespace;
+    this.deviceInterface!.switchToNamespace(namespace);
+    await this.pushNamespace(namespace);
+  }
+
+  private async createNamespace(isDefault: boolean = false, label?: string) {
```
mobitar, line 64: `isDefault: boolean = false` can be simplified to `isDefault = false`

lib/services/namespace_service.ts
```diff
@@ -0,0 +1,139 @@
...
+    if (isDefault) {
```
mobitar, line 65: Can be simplified to `if (isDefault && await this.getDefaultNamespace()) { ...exception...}`

lib/application.ts
```diff
@@ -1445,6 +1452,14 @@ export class SNApplication {
     this.services.push(this.actionsManager!);
   }
 
+  private createNamespaceService() {
+    this.namespaceService = new SNNamespaceService(this.deviceInterface!);
+    if (this.fixedNamespace) {
+      this.namespaceService!.createFixedNamespace(this.fixedNamespace);
```
mobitar, line 1458: I'm not yet sure how fixedNamespace functions, but sounds like this should be passed to the constructor instead, if it's something that shouldn't be called more than once.

lib/application.ts
```diff
@@ -78,7 +79,7 @@ export class SNApplication {
 
   public environment: Environment
   public platform: Platform
-  public namespace: string
+  private fixedNamespace?: string
```
mobitar, line 82: Seems like this should be called `namespaceIdentifier` instead? That's how it appears to be used. This should also have a comment in the application constructor as to why/when you would pass an identifier for a namespace.

mobitar: (On line 126)

lib/application.ts
```diff
@@ -179,6 +181,9 @@ export class SNApplication {
         return undefined;
       });
     this.createdNewDatabase = databaseResult?.isNewDatabase || false;
+    if (!this.fixedNamespace) {
+      await this.namespaceService!.initialize();
```
mobitar, line 185: `initialize` seems poorly named. It seems like it should be something like `assignNamespace`? Also, doesn't `deviceInterface!.openDatabase()` depend on the correct namespace being set? If so, seems like namespace initialization should be the first thing in prepareForLaunch.

lib/services/namespace_service.ts
```diff
@@ -0,0 +1,139 @@
...
+      namespace.isDefault = true;
+      namespace.label = 'Default namespace';
+      await this.setCurrentNamespace(namespace);
+    }
+    /**
+     * In the future, if a user has multiple accounts signed into a client, each account
+     * will use its own keychain/storage namespace. This service will be in charge of displaying
+     * a UI for selecting which namespace to sign into (if no default namespace is set).
+     */
+    else if (namespaces.length > 1) {
```
mobitar, line 111: Should just be `else { ... }` as there are no other conditions.

lib/services/namespace_service.ts
```diff
@@ -0,0 +1,139 @@
...
+      throw Error('Multiple namespaces not supported yet ;)')
+    }
+  }
+
+  /**
+   * Creates a namespace with a fixed identifier.
+   * @param identifier The namespace identifier.
+   */
+  public createFixedNamespace(identifier: string) {
```
mobitar, line 120: Seems like this function should just be grouped in with `createNamespace`, where `createNamespace` would hae an optional last parameter called `identifier`. Also, it's not clear why this function calls `setCurrentNamespace` but createNamespace doesn't.

lib/services/namespace_service.ts
```diff
@@ -0,0 +1,139 @@
...
+    const namespace: SNNamespace = {
+      identifier: identifier,
+      userUuid: undefined,
+      label: identifier,
+      isDefault: false
+    };
+    this.setCurrentNamespace(namespace);
```
mobitar, line 127: This is an async function but is not awaited. May cause race conditions.

lib/application.ts
johnny243: >initialize seems poorly named. It seems like it should be something like assignNamespace?

I was keeping the same standard, most of the services have an `initialize` method. In this case, this method just decides what namespace to use and switchs to it.

lib/application.ts
```diff
@@ -703,8 +710,8 @@ export class SNApplication {
   /**
    * Returns true if there is an encryption source available
    */
-  public isEncryptionAvailable() {
-    return this.hasAccount() || this.hasPasscode();
+  public async isEncryptionAvailable() {
+    return !isNullOrUndefined(this.getUser()) || this.hasPasscode();
```
mobitar, line 714: Why is this undoing changes @radko93 recently introduced?

radko93: Seems like a mistake

lib/services/namespace_service.ts
```diff
@@ -0,0 +1,131 @@
...
+    this.deviceInterface = deviceInterface;
+
+    if (namespaceIdentifier) {
+      const namespace: SNNamespace = {
+        identifier: namespaceIdentifier,
+        label: namespaceIdentifier,
+        isDefault: true
+      };
+      this.namespace = namespace;
+      this.deviceInterface!.switchToNamespace(namespace);
```
mobitar, line 33: This function will indirectly write to storage. We shouldn't be writing to storage in a constructor, since assumedly, nothing has changed. Is this assessment correct? If so, you need to separate what's inside `switchToNamespace` to extract the non-async part. You should never call an async function in a constructor, even if you don't need to await the result.

mobitar: Oh wait a minute. This is using deviceInterface.switchToNamespace. I thought it was using this class's function. In that case let me re-analyze.

lib/device_interface.ts
```diff
@@ -80,20 +79,19 @@ export abstract class DeviceInterface {
...
-  abstract async getKeychainValue() : Promise<any>;
+  abstract async clearNamespacedKeychainValue() : Promise<void>;
 
-  abstract async setKeychainValue(value: any) : Promise<void>;
+  abstract async getRawKeychainValue() : Promise<any>;
 
-  abstract async clearKeychainValue() : Promise<void>;
+  abstract async clearRawKeychainValue() : Promise<void>;
+
+  public switchToNamespace(namespace: SNNamespace) {
```
mobitar, line 92: This should just be called `setNamespace`, since it is quite literally a setter. `switchTo` implies there's heavy gears spinning to make the switch, but really it's just a setter.

lib/services/namespace_service.ts
```diff
@@ -0,0 +1,131 @@
...
+      await this.createNamespace(true);
+    }
+    /**
+     * If one (1) namespace exist, then it means it is not the default one at this point.
+     * Let's set it as the default namespace and return it.
+     */
+    else if (namespaces.length === 1) {
+      const namespace = namespaces[0];
+      namespace.isDefault = true;
+      namespace.label = 'Default namespace';
```
mobitar, line 95: "namespace" is an internal concept. For UI purposes, they should be called "applications". So this would be called "Default application"

lib/services/namespace_service.ts
```diff
@@ -0,0 +1,131 @@
...
+
+  private async createNamespace(isDefault = false, identifier?: string, label?: string) {
+    if (isDefault && await this.getDefaultNamespace()) {
+      throw Error('Can not create default namespace: a default namespace already exists.');
+    }
+    const uuid = await Uuid.GenerateUuid();
+    const namespace: SNNamespace = {
+      identifier: identifier || uuid,
+      userUuid: undefined,
+      label: isDefault ? 'Default namespace' : label || identifier || uuid,
```
mobitar, line 116: "Default application"

lib/application.ts
johnny243: Sorry about that, fixed :+1:

PR #57, opened 2020-07-30
feature: migrate passcode keyboard type

radko93:


lib/migrations/2020-01-15.ts
```diff
@@ -298,13 +299,18 @@ export class Migration20200115 extends Migration {
...
+      LegacyKeys.MobilePasscodeKeyboardTypeKey
+    );
     const rawStructure: StorageValuesObject = {
       [ValueModesKeys.Nonwrapped]: {
         [StorageKey.WrappedRootKey]: wrappedAccountKey,
         [StorageKey.RootKeyWrapperKeyParams]: rawPasscodeParams,
-        [StorageKey.RootKeyParams]: rawAccountKeyParams
+        [StorageKey.RootKeyParams]: rawAccountKeyParams,
+      },
+      [ValueModesKeys.Unwrapped]: {
```
radko93, line 311: What is the difference between `Unwrapped` and `Nonwrapped` in case of `ValueModesKeys`? Maybe it should be documented more explicitly. Because biometrics are set on `Unwrapped` and then we use `StorageValueModes.Nonwrapped` in the mobile repo to read it.

mobitar: Unwrapped means the value was wrapped previously (encrypted), then unwrapped and placed in working memory. Nonwrapped means the value should never be encrypted. So any values that are required to begin the unlocking process of the application should be nonwrapped (such as passcode keyboard type). This way we can access them before the user authenticates.

mobitar:
Shouldn't this key automatically be migrated by this function? https://github.com/standardnotes/snjs/blob/a27979314775f7aa34d4817bee740ab3fa3948c5/lib/migrations/2020-01-15.ts#L423

The aim of this function is to migrate any values that exist in raw storage (AsyncStorage) into the new managed storage service approach.

radko93:
@mobitar Oh, thanks. So I only added a new StorageKey and it should be possible to read it from nonwrapped storage?

mobitar:
I don't think you need to add StorageKey. It's an enum and it's never read as an array.
`migrateArbitraryRawStorageToManagedStorageAllPlatforms` uses `deviceInterface.getAllRawStorageKeyValues` to get all pre-existing raw storage values.

radko93:
Not for migration, but I should have a constant to use it later on. But I can create it on mobile then.

mobitar:
> Not for migration, but I should have a constant to use it later on. But I can create it on mobile then.

Can you rephrase 😅 No idea what you're referring to..

radko93:
I just need to store the key under a constant 😅

PR #58, opened 2020-07-30
feature: github action to check tsc on push

radko93:
Somehow running `npm run tsc` fails on the default branch. This motivated me to create a Github Action check so it's impossible to miss it next time when someone creates a PR. It runs `npm run tsc` on every push.


radko93:
@baptiste-grob do you want to me to copy this to the web as well?

baptiste-grob:
Not yet, I actually need to fix some typescript before :) thanks!

PR #61, opened 2020-08-03
feature: make alertService protected in Component Manager

radko93:
This enables classes that extend Component Manager to use alert service. Useful for `presentPermissionsDialog`.

### DIFF
--- a/dist/@types/services/component_manager.d.ts
+++ b/dist/@types/services/component_manager.d.ts
@@ -74,7 +74,7 @@ declare type ComponentState = {
 export declare class SNComponentManager extends PureService {
     private itemManager;
     private syncService;
-    private alertService;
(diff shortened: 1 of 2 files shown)


mobitar:
Can you elaborate more on your usecase? `presentPermissionsDialog` should be handled by SNJS and should not be a client concern..

radko93:
`presentPermissionsDialog` has to be overridden on the client-side https://github.com/standardnotes/snjs/blob/004/lib/services/component_manager.ts#L1165 so it would be nice to use `alertService`. The [web implementation](https://github.com/standardnotes/web/blob/develop/app/assets/javascripts/views/application/application_view.ts#L247) is not great either IMO I think ComponentManager should be subclassed.

PR #63, opened 2020-08-04
fix: safe app domain value access

radko93:
When signing in sometimes `appData` miss keys. This makes it safe to access any key.


mobitar:
Why would accessing a non-existent key be dangerous?

const foo = {};
const bar = foo['zoo'];
// bar = undefined, no exceptions thrown

It seems like the real safety check here is the question mark `appData?`? If so I don't think you need to use `hasOwnProperty`.

mobitar:
Btw access to appData was intentionally dangerous so that we catch programming issues. appData should really never be undefined as we fill it in when properly using the correct methods: https://github.com/standardnotes/snjs/blob/004/lib/models/functions.ts#L17

radko93:
Closing to fix the real issue instead.

PR #64, opened 2020-08-06
feature: properly sort encrypted items

radko93:
This ensures elements with `errorDecrypting` flag are safely put on top of the list.


lib/protocol/collection/item_collection.ts
```diff
@@ -118,22 +118,34 @@ export class ItemCollection extends MutableCollection<SNItem> {
       const filteredCTMap = this.filteredMap[contentType]!;
       const sortedElements = this.sortedMap[contentType]!;
       /** If the element is deleted, or if it no longer exists in the primary map (because
-       * it was discarded without neccessarily being marked as deleted), it does not pass 
+       * it was discarded without neccessarily being marked as deleted), it does not pass
        * the filter. If no filter the element passes by default. */
       const passes = (element.deleted || !this.map[element.uuid]) ? false : (filter ? filter(element) : true);
```
mobitar, line 123: So it would seem that the filter can also access properties that are meant to be decrypted, and could cause exception. I believe the `errorDecrypting` check should occur before we access the `filter` function.

Here is the filter function web sets: https://github.com/standardnotes/web/blob/develop/app/assets/javascripts/views/notes/notes_view.ts#L317

And this is the underlying helper function it uses: https://github.com/standardnotes/web/blob/1c3530befa6300cce66ef271a4f6806d4b4abb76/app/assets/javascripts/views/notes/note_utils.ts#L14

This function accesses .pinned and .archived, which should actually just be regular properties, so undefined if errorDecrypting. So I'm not sure if this would throw exception. So I'm curious what it was exactly that caused mobile to throw exception?

lib/protocol/collection/item_collection.ts
```diff
@@ -118,22 +118,34 @@ export class ItemCollection extends MutableCollection<SNItem> {
...
       const currentIndex = filteredCTMap[element.uuid];
       if (passes) {
         if (!isNullOrUndefined(currentIndex)) {
           /** Check to see if the element has changed its sort value. If so, we need to re-sort */
           const currentElement = sortedElements[currentIndex];
+          if (currentElement?.errorDecrypting) {
+            if (element.errorDecrypting) {
+             /** if both elements are decrypted skip compare */
```
mobitar, line 131: Do you mean "encrypted" here instead of decrypted?

lib/protocol/collection/item_collection.ts
radko93: `const previousValue = (currentElement as any)[sortBy.key];` this line invokes 
```
get userModifiedDate() {
    const value = this.getAppDomainValue(AppDataField.UserModifiedDate);
    return new Date(value || this.updated_at);
}
```
And this will crash on the encrypted item. I believe it's a race condition that this happens on mobile and does not on the web.

mobitar: You know what, getters that access decrypted content should be moved into a property instead. I started heading towards this direction but appear to have missed this for `userModifiedDate`. So you should add this after this line similar to other properties: https://github.com/standardnotes/snjs/blob/004/lib/models/core/item.ts#L89

mobitar: Because this way we are in the safety check: https://github.com/standardnotes/snjs/blob/004/lib/models/core/item.ts#L83

radko93: > So it would seem that the filter can also access properties that are meant to be decrypted, and could cause exception. I believe the errorDecrypting check should occur before we access the filter function.

I moved it to the top now.

lib/protocol/collection/item_collection.ts
```diff
@@ -117,11 +117,25 @@ export class ItemCollection extends MutableCollection<SNItem> {
...
       const filteredCTMap = this.filteredMap[contentType]!;
       const sortedElements = this.sortedMap[contentType]!;
+
+      const currentIndex = filteredCTMap[element.uuid];
+      const currentElement = !isNullOrUndefined(currentIndex) ? sortedElements[currentIndex] : undefined;
+      if (currentElement?.errorDecrypting) {
+        if (element.errorDecrypting) {
+         /** if both elements are encrypted skip compare */
+          sortedElements[currentIndex] = element;
+          continue;
```
mobitar, line 127: It looks like

```
          sortedElements[currentIndex] = element;
          continue;
```

is repeated in both if/else conditions, so can be factored out.

lib/protocol/collection/item_collection.ts
```diff
@@ -117,11 +117,25 @@ export class ItemCollection extends MutableCollection<SNItem> {
       /** Filtered content type map */
       const filteredCTMap = this.filteredMap[contentType]!;
       const sortedElements = this.sortedMap[contentType]!;
+
+      const currentIndex = filteredCTMap[element.uuid];
+      const currentElement = !isNullOrUndefined(currentIndex) ? sortedElements[currentIndex] : undefined;
```
mobitar, line 122: I think `previous` would make more sense instead of current when reading the code. `previousIndex` and `previousElement`.

lib/protocol/collection/item_collection.ts
```diff
@@ -117,11 +117,25 @@ export class ItemCollection extends MutableCollection<SNItem> {
...
+      if (currentElement?.errorDecrypting) {
+        if (element.errorDecrypting) {
+         /** if both elements are encrypted skip compare */
```
mobitar, line 125: Since they are the same elements (same UUID), this comment can be more helpful as:

```
/** If both previous and new values of element are errored, we simply replace the value, as a re-sort 
is not necessary, and the element will not be usable by forthcoming filters */
```

(Comments should use sentence case and so first letter should be capitalized)

radko93:
@mobitar I changed it a little bit again so it also works if the new element is encrypted.

lib/protocol/collection/item_collection.ts
```diff
@@ -117,23 +117,35 @@ export class ItemCollection extends MutableCollection<SNItem> {
       /** Filtered content type map */
       const filteredCTMap = this.filteredMap[contentType]!;
       const sortedElements = this.sortedMap[contentType]!;
+
+      const previousIndex = filteredCTMap[element.uuid];
+      const previousElement = !isNullOrUndefined(previousIndex) ? sortedElements[previousIndex] : undefined;
+      if (element.errorDecrypting || previousElement?.errorDecrypting) {
+        if (!element.errorDecrypting && previousElement?.errorDecrypting) {
```
mobitar, line 124: What if element is errorDecrypting, but new element is not errorDecrypting? Wouldn't we need to resort as well? In that case should this be something `element.errorDecrypting !== previousElement?.errorDecrypting`? (You'd need to account for potential of undefined on rhs).

lib/protocol/collection/item_collection.ts
```diff
@@ -117,23 +117,35 @@ export class ItemCollection extends MutableCollection<SNItem> {
...
+          /** If new element is not encrypted do a resort */
+         typesNeedingResort.add(contentType);
+        }
+         /** If both previous and new values of element are errored, we simply replace the value, as a re-sort
```
mobitar, line 128: This comment appears to no longer be valid since it's not addressing the case of both previous and new values being errored.

lib/protocol/collection/item_collection.ts
radko93: I did like this now. In case old value is undefined we just don't need to care about it (new element with `true` value will pass this check).

lib/protocol/collection/item_collection.ts
```diff
@@ -117,23 +117,36 @@ export class ItemCollection extends MutableCollection<SNItem> {
...
+      const previousElement = !isNullOrUndefined(previousIndex) ? sortedElements[previousIndex] : undefined;
+      if (element.errorDecrypting || previousElement?.errorDecrypting) {
+        if (element.errorDecrypting !== previousElement?.errorDecrypting) {
+          /** If errorDecrypting property changes a re-sort is necessary.
+           * If previous value is undefined we don't need to care about it. */
+         typesNeedingResort.add(contentType);
+        }
+         /** If the errorDecrypting property stayed the same, we simply replace the value, as a re-sort
+          is not necessary, and the element will not be usable by forthcoming filters */
+        sortedElements[previousIndex] = element;
```
mobitar, line 131: Wait, if new `element` is **not** errored, doesn't that mean it is valid, and we can use it in filters below?

radko93: I update this to check if the new element is valid. Also added a matrix of all possibilities because there are so many options here :D

lib/protocol/collection/item_collection.ts
```diff
@@ -117,23 +117,46 @@ export class ItemCollection extends MutableCollection<SNItem> {
...
+      if (element.errorDecrypting || previousElement?.errorDecrypting) {
+        if (element.errorDecrypting !== previousElement?.errorDecrypting) {
+          /** If errorDecrypting property changes a re-sort is necessary. */
+         typesNeedingResort.add(contentType);
+        }
+         /** We always save the value, but we don't continue in the for loop
+          * in case the errorDecrypting is false in the new value as we want filters to check it.  */
+        sortedElements[previousIndex] = element;
+        if (!element.errorDecrypting) {
+          continue;
```
mobitar, line 140: Wait why would you continue if the element is not error decrypting? If it is valid, then we should proceed to filtering. If it is invalid, we should skip filtering.

However, come to think about it, even if an item is invalid, we don't want to skip filtering, because the [filter is how clients choose](https://github.com/standardnotes/web/blob/004/app/assets/javascripts/views/notes/notes_view.ts#L320) which notes to display based on active tag. In this case it would seem like we would always display errorDecrypting items for every tag a user selects.

I'm thinking that perhaps clients should be the ones to do safety check for errorDecrypting. Perhaps the client wants to display a list of items that are errorDecrypting, sorted by created_at, or something like that? In which case, I think we could scrap this PR in its entirety.

radko93: Yeah, that might be true. I removed most of the logic, but I left safe emergency safety checks and the test so IMO we should merge this.

PR #66, opened 2020-08-06
refactor: safe userModifiedDate

radko93:
I'm not sure if it's best to keep this value optional or use `updated_at` (as it was default before). Also not sure about setter yet.


lib/models/core/item.ts
```diff
@@ -81,12 +82,15 @@ export class SNItem {
     this.conflictOf = payload.safeContent.conflict_of;
     this.createdAtString = this.created_at && this.dateToLocalizedString(this.created_at);
     if (payload.format === PayloadFormat.DecryptedBareObject) {
+      this.userModifiedDate = this.getAppDomainValue(AppDataField.UserModifiedDate);
```
mobitar, line 85: Shouldn't this be wrapped in a Date constructor? Should be 

```javascript
this.userModifiedDate = new Date(this.getAppDomainValue(AppDataField.UserModifiedDate) || this.updated_at)`
```

mobitar: Otherwise everything else is correct I think.

PR #67, opened 2020-08-07
feature: mobile user preferences

radko93:
This brings mobile-specific preferences to snjs. I wonder if pref keys should be different from the web.

### DIFF
--- a/lib/models/app/userPrefs.ts
+++ b/lib/models/app/userPrefs.ts
@@ -17,6 +17,20 @@ export enum WebPrefKey {
   NotesHideDate = 'hideDate'
 };
 
+export enum MobilePrefKey {
+  SortNotesBy = 'sortBy',
+  SortNotesReverse = 'sortReverse',
+  NotesShowArchived = 'showArchived',
+  NotesHidePinned = 'hidePinned',
+  NotesHideNotePreview = 'hideNotePreview',
+  NotesHideDate = 'hideDate',
+  ThemeData = 'themePreferences',
+  LastExportDate = 'lastExportDate',
+  DoNotWarnUnsupportedEditors = 'doNotShowAgainUnsupportedEditors'
+}
+
+export type PrefKey = WebPrefKey | MobilePrefKey
+
 export class SNUserPrefs extends SNItem {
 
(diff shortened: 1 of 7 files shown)


lib/models/app/userPrefs.ts
```diff
@@ -36,4 +50,7 @@ export class UserPrefsMutator extends ItemMutator {
   setWebPref(key: WebPrefKey, value: any) {
     this.setAppDataItem(key as any, value);
   }
+  setMobilePref(key: MobilePrefKey, value: any) {
+    this.setAppDataItem(key as any, value);
```
mobitar, line 54: Since, for example, `sortBy` is present in both `MobilePrefKey` and `WebPrefKey`, wouldn't that mean setting that key on mobile would overwrite the web preference?

radko93: Yes indeed. I need to use different keys then.

baptiste-grob: If value clashing is something we have to avoid then I think we should merge `MobilePrefKey` and `WebPrefKey` to have the compiler make sure there is no overlap.

mobitar: I guess it's tradeoffs. I like the merge idea but also keeping them separate seems..cleaner?

mobitar:
Does a user's old sort by settings, etc, get migrated into this new format?

radko93:
@mobitar no, I need to write migration as well separately.

lib/models/app/userPrefs.ts
baptiste-grob: I agree that it makes things clearer, however merging the enums will prevent overlap bugs, since they will be caught at compile time. IMO this is the winning tradeoff but maybe there's another way to have compile-time safety

mobitar: What if you create an unused enum that is the intersection of both, that simply exists for type safety (if that is even possible)?

baptiste-grob: I'm thinking and I don't think there's another decent way to model this…

PR #68, opened 2020-08-11
feature: default host for application

radko93:
This makes it possible to set a default host when creating an application (so it does not have to be set manually).


lib/services/api/api_service.ts
```diff
@@ -49,7 +50,7 @@ export class SNApiService extends PureService {
 
   public async loadHost() {
     const storedValue = await this.storageService!.getValue(StorageKey.ServerHost);
-    this.host = storedValue || (window as any)._default_sync_server;
+    this.host = storedValue || this.host || (window as any)._default_sync_server;
```
radko93, line 53: @baptiste-grob I left `(window as any)._default_sync_server;` for now but it should be removed when web adapts. Or maybe you want to remove it now?

baptiste-grob: We can already remove it yes

PR #74, opened 2020-08-17
feature: expose toggleComponent method

baptiste-grob:
So that clients don't have to access the componentManager


mobitar:
Function could use a description though..

baptiste-grob:
I added a description. Let me know what you think.

PR #75, opened 2020-08-17
Actually clear all data

baptiste-grob:
Fixes: StorageService.clearAllData doesn't actually clear all data and can be invalidated by subsequent calls to removeValue.

### DIFF
--- a/lib/services/storage_service.ts
+++ b/lib/services/storage_service.ts
@@ -363,10 +363,11 @@ export class SNStorageService extends PureService {
     });
   }
 
-  public async clearAllData() {
-    return Promise.all([
-      this.clearValues(),
-      this.clearAllPayloads()
-    ]);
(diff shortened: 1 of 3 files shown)


lib/services/storage_service.ts
```diff
@@ -360,10 +363,11 @@ export class SNStorageService extends PureService {
...
-  public async clearAllData() {
-    return Promise.all([
-      this.clearValues(),
-      this.clearAllPayloads()
-    ]);
+  public clearAllData(): Promise<void> {
+    return this.executeCriticalFunction(async () => {
+      await this.clearValues();
+      await this.clearAllPayloads();
+      await this.deviceInterface!.removeAllRawStorageValues();
```
mobitar, line 370: This might interfere with namespacing. When you sign out of one account, we don't want to sign out of all accounts.

PR #76, opened 2020-08-18
fix: importing deleted items

johnny243:
- [x] fix the payload source that is overwritten and set to `PayloadSource.Constructor` if no `source` parameter is passed to [CreatePayload](https://github.com/standardnotes/snjs/blob/a8205e4e9a3b74815c7371230481ea24461c0c40/lib/protocol/payloads/generator.ts#L253)
- [x] new test [for previously deleted items](https://github.com/standardnotes/snjs/blob/a8205e4e9a3b74815c7371230481ea24461c0c40/test/model_tests/importing.test.js#L243) that uses encrypted data
- [x] new tests to check imported item's PayloadSource


lib/protocol/payloads/generator.ts
```diff
@@ -268,7 +268,7 @@ function CreatePayload(
   return new PurePayload(
     rawPayload,
     newFields,
-    source || PayloadSource.Constructor
+    source || object.source || PayloadSource.Constructor
```
mobitar, line 271: Hmmmm...can you show which line of code will call this function with object.source being valid? And in that same location, why not just pass the source instead?

johnny243: https://github.com/standardnotes/snjs/blob/a8205e4e9a3b74815c7371230481ea24461c0c40/lib/services/protocol_service.ts#L540

[CreateEncryptionParameters](https://github.com/standardnotes/snjs/blob/a8205e4e9a3b74815c7371230481ea24461c0c40/lib/protocol/payloads/generator.ts#L275) is calling [CreatePayload](https://github.com/standardnotes/snjs/blob/a8205e4e9a3b74815c7371230481ea24461c0c40/lib/protocol/payloads/generator.ts#L253).

> And in that same location, why not just pass the source instead?

I could add an optional parameter `source` in [CreateEncryptionParameters](https://github.com/standardnotes/snjs/blob/a8205e4e9a3b74815c7371230481ea24461c0c40/lib/protocol/payloads/generator.ts#L275). But I think [CreatePayload](https://github.com/standardnotes/snjs/blob/a8205e4e9a3b74815c7371230481ea24461c0c40/lib/protocol/payloads/generator.ts#L253) should use the payload `source` (if any). If not fallback to the `Constructor` source.

mobitar: I dunno, I'm torn on this. `object` is `any` type, so we can't assume it has a `source` property. Sure, it's safe and won't cause exceptions, but it does create code confusion. I think we should always try to be more explicit than implicit. So adding a `source` parameter to `CreateEncryptionParameters` is much more explicit and cleaner overall, and much easier to track, if a developer one year from now were reading this code.

PR #80, opened 2020-08-18
feature: expose getAsboluteUrl from application

radko93:
This PR makes it possible to use http service by an application (to download themes on mobile). This could alternatively be in `MobileApplication`?


mobitar:
Hmm..this is too weird and dangling. Perhaps just create your own instance of HttpService in the mobile app?

PR #82, opened 2020-08-19
fix: session history in descending order

johnny243:
- [x] entries are now sorted in descending order


lib/services/history/session/item_session_history.ts
```diff
@@ -21,20 +20,21 @@ export class ItemSessionHistory {
...
       return CreateHistoryEntryForPayload(rawHistoryEntry.payload);
-    })
+    });
     return new ItemSessionHistory(entries);
   }
 
   getLastEntry() {
-    return this.entries[this.entries.length - 1];
+    /** First element in the array should be the last entry. */
+    return this.entries[0];
```
mobitar, line 37: This function should be renamed entirely to "getMostRecentEntry"

mobitar: "last" is strictly defined ;)

PR #83, opened 2020-08-20
fix: make removePrivatePropertiesFromResponseItems handle immutable data 

baptiste-grob:
ComponentManager.removePrivatePropertiesFromResponseItems used to mutate data that is now read-only. This PR changes the method so it returns new data instead or altering its inputs.


mobitar:
I do like the approach of returning new objects instead of modifying inputs. But I'm not sure where the read only data was? It seems like everything passed into the function is mutable?

PR #86, opened 2020-08-23
fix: remove multilines from messages

radko93:


mobitar:
@radko93 there is an extraneous space here: https://github.com/standardnotes/snjs/blob/004/lib/services/api/messages.ts#L43

> is_reporting_a__version

PR #87, opened 2020-08-25
fix: extension visibility should not be persisted

johnny243:
No need to persist extension visibility.


baptiste-grob:
We do need `hidden` to be persisted for the components that sit in the bottom stack (Action Bar, FileSafe etc…) so the values stay the same between clients. When you say extension are you referring to actions specifically?

mobitar:
This property appears to apply only to Action extensions. We already migrated hidden state management to be transient in SNJS component_manager.

However, I think we should keep persistence here. If a user has many Action extensions and they collapse a certain section in the menu, this collapsed state should be remembered. This was actually an explicit feature request in the past.

PR #90, opened 2020-08-26
display blocking dialog during passcode operations

baptiste-grob:
Passcode operations can take some time. This change warns the user not to close the app during these destructive operations.


baptiste-grob:
(Haven't run tests yet, will request a review after)

PR #92, opened 2020-08-27
fix: do not activate or deactivate editors

baptiste-grob:
It doesn't make sense to track an editor's `active/inactive` state but before this change it was still happening and could cause bugs when dealing with existing data.

@radko93 If you're using activate/deactivateComponent with editors I don't think it would cause an issue but you can't rely on their active/inactive state and should instead use ComponentManager.editorForNote(SNNote) to determine if an editor should be displayed

### DIFF
--- a/lib/services/component_manager.ts
+++ b/lib/services/component_manager.ts
@@ -223,6 +223,9 @@ export class SNComponentManager extends PureService {
           }
         }
         for (const component of syncedComponents) {
+          if (component.isEditor()) {
+            continue;
+          }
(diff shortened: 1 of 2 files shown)


lib/services/component_manager.ts
```diff
@@ -223,6 +223,9 @@ export class SNComponentManager extends PureService {
           }
         }
         for (const component of syncedComponents) {
+          if (component.isEditor()) {
```
mobitar, line 226: Might it be worth a comment here? Kind of an abrupt condition without the proper context..

PR #96, opened 2020-08-29
fix: session history tests

johnny243:
- [x] test session history entries order (should be from newest to oldest)
- [x] fix test "should optimize basic entries"
- [x] fix test "should keep the entry right before a large deletion, regardless of its delta"

### DIFF
--- a/dist/snjs.js
+++ b/dist/snjs.js
@@ -17695,7 +17695,7 @@ class item_session_history_ItemSessionHistory {
 
       if (keep && isEntrySignificant(entry) && entry.operationVector() === -1) {
         /** This is a large negative change. Hang on to the previous entry. */
-        const previousEntry = this.entries[index - 1];
+        const previousEntry = this.entries[index + 1];
 
         if (previousEntry) {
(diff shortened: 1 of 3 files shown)


johnny243:
@mobitar I believe both ["should optimize basic entries"](https://github.com/standardnotes/snjs/blob/e5329c066c7eb853c5ee405ed8f11ff34649f604/test/history.test.js#L85) and ["should keep the entry right before a large deletion, regardless of its delta"](https://github.com/standardnotes/snjs/blob/e5329c066c7eb853c5ee405ed8f11ff34649f604/test/history.test.js#L159) need to be updated according to https://github.com/standardnotes/snjs/commit/2e6aa1225385dd6a48d810ff7259fa27ff44429c, these tests are failing due to this commit.

mobitar:
Yes if the sort order has changed these tests would need to be updated. Small fix it would seem. "Keep the entry right before" probably means it was using index - 1 to get the before entry, but in the new code, it will be index + 1.

PR #99, opened 2020-09-01
fix: enforce correct type for getAllRawStorageKeyValues

radko93:
Because we've used `any` it was possible to return the wrong data format in `getAllRawStorageKeyValues`. This the type which web device interface actually returns and the one that is used in migrations.

### DIFF
--- a/dist/@types/device_interface.d.ts
+++ b/dist/@types/device_interface.d.ts
@@ -28,7 +28,10 @@ export declare abstract class DeviceInterface {
      * So we return the value as-is if JSON.parse throws an exception.
      */
     getJsonParsedStorageValue(key: string): Promise<any>;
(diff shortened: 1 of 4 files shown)


mobitar:
Not sure I'm following..why is `unknown` any better than `any`?

radko93:
This is actually not really about `unknown` but about enforcing `key` + `value` records format.`unknown` provides a bit more type safety in general.

baptiste-grob:
You can't do anything on `unknown` types, they are considered to have no method or value on them, essentially saying the code has to figure it out itself. At the very least you have to cast it to another type. With `any`, anything goes, you can try to access whatever member on it or pass it to any function you want and the compiler won't care

baptiste-grob (approved):

PR #101, opened 2020-09-07
fix: handle case where user may have 003 key not set as default

mobitar:
I'm not sure how my 003 account got into a state where it had a single itemsKey, but isDefault was not set to true. So upon sign in to the account, when the application attempted to persist data locally, it searched for any key where isDefault is true, and that key was an ephemeral 004 key that the application created on launch.

This ephemeral items key still existed because the full sync had not completed yet, otherwise, it would have been deleted at that point. But because we persist to storage incrementally, it was still a valid key available to SNJS.

It seems like we should be able to perform [this](https://github.com/standardnotes/snjs/blob/004/lib/services/protocol_service.ts#L1230) step on incremental sync as well, not just full syncs. The reason I'm applying this as a mostly corrective fix for now is that I'm not sure how my account got into this state and it may very well be due to fluxes in the development environment. If we find that the console.warn added in this PR occurs often, we'll need to investigate how accounts get into that state.

### DIFF
--- a/lib/services/protocol_service.ts
+++ b/lib/services/protocol_service.ts
@@ -1273,7 +1273,7 @@ export class SNProtocolService extends PureService implements EncryptionDelegate
 
   private async handleFullSyncCompletion() {
     /** Always create a new items key after full sync, if no items key is found */
-    const currentItemsKey = this.getDefaultItemsKey();
+    const currentItemsKey = await this.getDefaultItemsKey();
     if (!currentItemsKey) {
       await this.createNewDefaultItemsKey();
       if (this.keyMode === KeyMode.WrapperOnly) {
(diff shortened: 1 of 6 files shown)


lib/services/protocol_service.ts
```diff
@@ -1311,14 +1311,27 @@ export class SNProtocolService extends PureService implements EncryptionDelegate
...
-    return itemsKeys.find((key) => {
+    const defaultKey = itemsKeys.find((key) => {
       return key.isDefault;
     });
+
+    /**
+     * The default key appears to be either newer or older than the user's account version
+     * We could throw an exception here, but will instead fall back to a corrective action:
+     * return any items key that corresponds to the user's version
+     */
```
baptiste-grob, line 1327: This comment should be inside the `if` block

PR #102, opened 2020-09-08
Backup file version

baptiste-grob:


baptiste-grob:
@radko93 This changes the signature of `importData` (it can now return an error string)

mobitar (approved):

PR #103, opened 2020-09-09
Fix/smart tags

baptiste-grob:
Introduces the `NotesCollection` class, which is essentially a specialized view into ItemCollection, which allows us to keep the ItemManager’s collection as the canonical source of items while also letting us apply specialized, potentially more performant filtering. In this case we solves the problem where Smart Tags have to check on a note’s tags (a property not present by default on a note) and optimize system tags by short-circuiting into custom logic.

This change is also backwards-compatible but @radko93 you may still want to use SNApplication.setNotesDisplayOptions instead of sorting tags & smart tags manually.


lib/models/core/predicate.ts
```diff
@@ -1,11 +1,21 @@
 import { SNItem } from '@Models/core/item';
 import { isString } from '@Lib/utils';
+import { version } from 'punycode';
```
mobitar, line 3: Lol what is punycode?

lib/protocol/collection/notes_collection.ts
```diff
@@ -0,0 +1,94 @@
...
+import { ItemCollection, CollectionSort, SortDirection } from "./item_collection";
+import { UuidString } from "@Lib/types";
+
+export class NotesCollection {
+  private displayedList: SNNote[] = [];
+  private tag?: SNTag;
+  needsRebuilding = true;
+
+  constructor(private collection: ItemCollection) {
+  }
```
mobitar, line 11: Not retaining passed in collection object?

lib/protocol/collection/notes_collection.ts
```diff
@@ -0,0 +1,94 @@
+import { SNNote, SNTag, SNSmartTag, SNPredicate, SNItem, ContentType } from '../../models'
+import { ItemCollection, CollectionSort, SortDirection } from "./item_collection";
+import { UuidString } from "@Lib/types";
+
+export class NotesCollection {
```
mobitar, line 5: Why did you decide against subclassing ItemCollection?

lib/protocol/collection/notes_collection.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  constructor(private collection: ItemCollection) {
+  }
+
+  private tags() {
+    return (this.collection.typedMap[ContentType.Tag] || []) as SNTag[];
+  }
+
+  public notesMatchingSmartTag(smartTag: SNSmartTag, notes: SNNote[]): SNNote[] {
+    let predicate = smartTag.predicate;
```
mobitar, line 18: Should be const?

lib/protocol/collection/notes_collection.ts
```diff
@@ -0,0 +1,94 @@
...
+      return matchingNotes as SNNote[];
+    }
+
+    if (predicate.keypathIncludesVerb('tags')) {
+      /** Populate notes with their tags */
+      const tags = this.tags();
+      matchingNotes = notes.map(note => ({
+        ...note,
+        ...note.payload,
+        tags: tags.filter(tag => tag.hasRelationshipWithItem(note))
```
mobitar, line 39: 1. What exactly is `tags` used for? After this condition you end up returning `this.collection.map[note.uuid]` and so the custom object you created here appears to be ignored?

2. What exactly is the type of `{...note, ...note.payload}`? Seems like an awkward concatenation. What is the purpose of combining them?

lib/services/item_manager.ts
```diff
@@ -59,15 +60,26 @@ export class ItemManager extends PureService {
...
     this.createCollection();
     this.unsubChangeObserver = this.modelManager
-      .addObserver(ContentType.Any, this.onPayloadChange.bind(this));
-    this.systemSmartTags = BuildSmartTags();
+      .addObserver(ContentType.Any, this.setPayloads.bind(this));
+  }
+
+  private createCollection() {
+    this.collection = new ItemCollection();
+    this.collection.setDisplayOptions(ContentType.Note, CollectionSort.CreatedAt, 'dsc');
```
mobitar, line 77: Is this line still necessary?

lib/services/item_manager.ts
```diff
@@ -241,6 +241,7 @@ export class ItemManager extends PureService {
     for (const item of discardedItems) {
       this.collection.discard(item);
     }
+    this.notesCollection.needsRebuilding = true;
```
mobitar, line 244: I feel like the collection itself should be smart enough to know when it needs rebuilding based on when its own properties change, rather than an external source telling it to rebuild.

lib/models/core/predicate.ts
baptiste-grob: …What the

lib/protocol/collection/notes_collection.ts
baptiste-grob: When you specify the modifier in the constructor it auto-assigns it: https://www.typescriptlang.org/play?#code/MYGwhgzhAECC0G8BQ1rAPYDsIBcBOArsDungBQAOeAlgG5g4Cm0ARgFzSYEC2LjeASgAFCAL5JRQA

lib/protocol/collection/notes_collection.ts
baptiste-grob: ItemCollection handles multiple content types. To subclass it would have meant that NotesCollection would be expected to handle arbitrary content types, which doesn't make sense and would likely lead to misunderstandings down the road. 
By using composition (NotesCollection has an ItemCollection) NotesCollection becomes a specialized view of ItemCollection, which makes it easier to pull out or change, and easier to understand (no implicit parent behavior).

lib/protocol/collection/notes_collection.ts
baptiste-grob: 1. `tags` is there so that existing predicates like `"tags.length"` can work.

2. Because I'm mapping to a plain object we lose every note getters, so I put the payload properties at the top-level to support other predicates. 
But you're right, it is kind of awkward. I'll get back to you with another approach.

lib/protocol/collection/notes_collection.ts
baptiste-grob: Maybe this needs another name actually, like ItemCollectionNotesView.

lib/services/item_manager.ts
baptiste-grob: I ruled out subclassing as I explained in another comment. Maybe I do need to rename this class to avoid the confusion, so that an external source telling it to rebuild wouldn't seem strange.

lib/services/item_manager.ts
baptiste-grob: It is. The NotesCollection leaves sorting to the ItemCollection and only does filtering

baptiste-grob: …Yeah the name definitely needs to change

lib/protocol/collection/notes_collection.ts
baptiste-grob: Other alternatives I considered were:
1) to have ItemCollection include NotesCollection and leave all the sorting and filtering to it. I was worried of bloating the itemCollection class with special cases however.
2) to have NotesCollection be a MutableCollection subclass and handle sorting itself. This would involve extracting the sorting functionality out of ItemCollection and into generic functions that NotesCollection could call into, and not feeding ItemCollection with notes anymore, to prevent unnecessary computations.

I was strongly tempted by 2) and the reason I went with my current implementation is that it seemed less disruptive and kept one single source of truth vs two (NotesCollection for notes vs ItemCollection for everything else).

lib/protocol/collection/item_collection_notes_view.ts
```diff
@@ -0,0 +1,105 @@
...
+  public notesMatchingSmartTag(smartTag: SNSmartTag, notes: SNNote[]): SNNote[] {
+    const predicate = smartTag.predicate;
+
+    /** Optimized special cases */
+    if (smartTag.isArchiveTag) {
+      return notes.filter(note => note.archived && !note.trashed && !note.deleted);
+    } else if (smartTag.isTrashTag) {
+      return notes.filter(note => note.trashed && !note.deleted);
+    }
+    const allNotes: SNNote[] = notes.filter(
```
mobitar, line 24: Is `SNNote[] ` definition still required here?

lib/protocol/collection/item_collection_notes_view.ts
```diff
@@ -0,0 +1,105 @@
...
+      note => !note.trashed && !note.deleted
+    );
+    if (smartTag.isAllTag) {
+      return allNotes as SNNote[];
```
mobitar, line 28: Is SNNote[] definition still required here?

mobitar (approved):

PR #106, opened 2020-09-10
fix: re-queue items for singleton processing when their encryption st…

mobitar:
…atus has changed

### DIFF
--- a/dist/@types/models/core/item.d.ts
+++ b/dist/@types/models/core/item.d.ts
@@ -150,6 +150,7 @@ export declare class ItemMutator {
     setDeleted(): void;
     set lastSyncBegan(began: Date);
     set errorDecrypting(errorDecrypting: boolean);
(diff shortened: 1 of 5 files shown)


dist/@types/models/core/item.d.ts
```diff
@@ -150,6 +150,7 @@ export declare class ItemMutator {
     setDeleted(): void;
     set lastSyncBegan(began: Date);
     set errorDecrypting(errorDecrypting: boolean);
+    set errorDecryptingValueChanged(errorDecryptingValueChanged: boolean);
```
baptiste-grob, line 153: Is there no other way to test this than to make a test-specific setter? When else would you use this?

baptiste-grob (approved):

dist/@types/models/core/item.d.ts
mobitar: Yeah you're right, there should be another way to do this without adding a new otherwise unused function.

PR #107, opened 2020-09-10
fix: prevent importing new backup file into old account

baptiste-grob:
We already prevent importing a backup from an unsupported protocol version, but we also need to prevent importing backups from known version into older accounts.

### DIFF
--- a/lib/application.ts
+++ b/lib/application.ts
@@ -51,13 +51,14 @@ import {
   SETTING_PASSCODE,
   REMOVING_PASSCODE,
   CHANGING_PASSCODE,
-  UNSUPPORTED_BACKUP_FILE_VERSION,
-  DO_NOT_CLOSE_APPLICATION
+  BACKUP_FILE_MORE_RECENT_THAN_ACCOUNT,
+  DO_NOT_CLOSE_APPLICATION,
+  UNSUPPORTED_BACKUP_FILE_VERSION
 } from './services/api/messages';
 import { MINIMUM_PASSWORD_LENGTH } from './services/api/session_manager';
 import { SNNamespaceService } from '@Services/namespace_service';
 import { SNComponent, SNTag, SNNote } from './models';
-import { ProtocolVersion } from './protocol/versions';
+import { ProtocolVersion, compareVersions } from './protocol/versions';
 
 
 /** How often to automatically sync, in milliseconds */
@@ -708,7 +709,7 @@ export class SNApplication {
     return this.protocolService!.getDefaultOperatorEncryptionDisplayName();
   }
 
-  public async getUserVersion() {
+  public getUserVersion() {
     return this.protocolService!.getUserVersion();
   }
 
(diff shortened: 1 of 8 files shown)


lib/application.ts
```diff
@@ -708,7 +709,7 @@ export class SNApplication {
     return this.protocolService!.getDefaultOperatorEncryptionDisplayName();
   }
 
-  public async getUserVersion() {
+  public getUserVersion() {
```
mobitar, line 712: Wait why remove `async` here? `getUserVersion` is async.

baptiste-grob: An async function is just a function that returns a promise. When we say:
```ts
public async getUserVersion() {
  return this.protocolService!.getUserVersion();
}
```

We're saying
```ts
public getUserVersion() {
  return new Promise(resolve => {
    this.protocolService!.getUserVersion().then(resolve);
  });
}
```

This is redundant and potentially costs us an extra event loop cycle (I'm not 100% but I believe the spec dictates Promises resolve on the next tick).
The async modifier should thus only be used for functions that need to use the await keyword. You can absolutely call an async function inside a non-async function and treat its result like the promise that it it, you just can't use await on that promise.

mobitar: Hmm..I usually read code as if it were documentation sometimes, because really we don't have docs for SNJS. Very often I check a function's signature to see how I should call it. Here I would probably look at this and say, oh cool, this function is sync, and I would call it as such in a unit test (where we don't have TS).

baptiste-grob: How about this then?
```
public getUserVersion(): Promise<ProtocolVersion | undefined> {
  return this.protocolService!.getUserVersion();
}
```

mobitar: I suppose that works..

lib/application.ts
```diff
@@ -811,10 +812,18 @@ export class SNApplication {
        * stop importing if there is no backup file version, only if there is
        * an unsupported version.
        */
+      const version = data.version as ProtocolVersion;
+
       const supportedVersions = this.protocolService!.supportedVersions();
-      if (!supportedVersions.includes(data.version as ProtocolVersion)) {
+      if (!supportedVersions.includes(version)) {
```
mobitar, line 818: Do you want to check if version can be undefined here? Similar to the hotfix yesterday.

lib/application.ts
```diff
@@ -811,10 +812,18 @@ export class SNApplication {
...
         return { error: UNSUPPORTED_BACKUP_FILE_VERSION };
       }
+
+      const userVersion = await this.getUserVersion();
+      if (userVersion && compareVersions(version, userVersion) === 1) {
```
mobitar, line 823: Same check here, version can be undefined.

lib/application.ts
baptiste-grob: This check is done before entering the block https://github.com/standardnotes/snjs/pull/107/files/7b9c188e7fac16ea14f7018f9bc1cd22f3ab9579#diff-958a601413ee4d47cb1165fe9ed67ffcL808

lib/application.ts
baptiste-grob: https://github.com/standardnotes/snjs/pull/107/files/7b9c188e7fac16ea14f7018f9bc1cd22f3ab9579#diff-958a601413ee4d47cb1165fe9ed67ffcL808

lib/application.ts
mobitar: Ah right 👍

mobitar (approved):

PR #109, opened 2020-09-11
feat: application group

mobitar:
Refactors namespace service into application group.

Application group is partially extracted from the web app.

Application groups manage multiple application instances.

Related: https://github.com/standardnotes/web/pull/468


johnny243 (approved):

johnny243 (commented):
Looks good! I know UI/UX will change in the future, but can we make the application name clickable? Currently, you have to click the area around the title in order to switch accounts.

mobitar:
@johnny243 do you mean this for the web PR? I'll double check that.

johnny243:
> @johnny243 do you mean this for the web PR? I'll double check that.

Yes, this is for the web PR. I accidentally commented here thinking I was on web. I'll provide more details in the correct PR.

lib/application_group.ts
```diff
@@ -0,0 +1,224 @@
+import { RawStorageKey } from '@Lib/storage_keys';
+import { removeFromArray, findInArray } from '@Lib/utils';
+import { UuidString, DeinitSource } from './types';
+import { SNApplication } from './application';
+import { PureService } from '@Services/pure_service';
+import { DeviceInterface } from '@Lib/device_interface';
+import { Uuid } from '@Lib/uuid';
+
+export type ApplicationDescriptor = {
+  identifier: string | UuidString,
```
baptiste-grob, line 10: Why not just a `UuidString`?

lib/application_group.ts
```diff
@@ -0,0 +1,224 @@
...
+  // userUuid?: UuidString,
```
baptiste-grob, line 11: Who's this little one?

lib/application_group.ts
```diff
@@ -0,0 +1,224 @@
...
+    super.deinit();
+    this.deviceInterface.deinit();
+    (this.deviceInterface as any) = undefined;
+  }
+
+  public async initialize(callback: AppGroupCallback) {
+    this.callback = callback;
+    this.descriptorRecord = (
+      await this.deviceInterface!.getJsonParsedRawStorageValue(RawStorageKey.DescriptorRecord)
+    ) as DescriptorRecord;
```
baptiste-grob, line 49: Can we trust this value enough to cast it? What would happen if it wasn't formatted in the way we expect?

lib/application_group.ts
```diff
@@ -0,0 +1,224 @@
...
+  private async createDescriptorRecord() {
+    const descriptorRecord = {} as DescriptorRecord;
+    const descriptor = {
+      /** The identifier 'standardnotes' is used because this was the database name of
+       * Standard Notes web/desktop */
+      identifier: "standardnotes",
+      label: 'Main Application',
+      primary: true
+    } as ApplicationDescriptor;
+    descriptorRecord[descriptor.identifier] = descriptor;
```
baptiste-grob, line 71: why not initialize descriptorRecord here so you don't have to cast it (destructive operation)? i-e:
```ts
const descriptorRecord: DescriptorRecord = { [descriptor.identifier]: descriptor };
```

lib/application_group.ts
```diff
@@ -0,0 +1,224 @@
...
+    this.descriptorRecord = descriptorRecord;
+    this.persistDescriptors();
+  }
+
+  public getApplications() {
+    return this.applications;
+  }
+
+  public getDescriptors() {
+    return Object.keys(this.descriptorRecord).map(key => this.descriptorRecord[key]!);
```
baptiste-grob, line 85: Should be
```ts
return Object.values(this.descriptorRecord);
```

lib/application_group.ts
```diff
@@ -0,0 +1,224 @@
...
+  }
+
+  private findPrimaryDescriptor() {
+    for (const key of Object.keys(this.descriptorRecord)) {
```
baptiste-grob, line 89: You could use `this.getDescriptors()` here as well

lib/application_group.ts
```diff
@@ -0,0 +1,224 @@
...
+    return this.descriptorRecord[application.identifier];
+  }
+
+  public async addNewApplication(label?: string) {
+    const identifier = await Uuid.GenerateUuid();
+    const index = this.getDescriptors().length + 1;
+    const descriptor = {
+      identifier: identifier,
+      label: label || `Application ${index}`
+    } as ApplicationDescriptor;
```
baptiste-grob, line 198: should be
```ts
const descriptor: ApplicationDescriptor = { ... }
```
Instead of casting. That way if we ever change ApplicationDescriptor we'll have errors pointing to places where we need to update code

baptiste-grob (approved):
Where'd dist/snjs.js go? 😄

lib/application_group.ts
johnny243: I believe `string` is for the default identifier: `standardnotes` and `UuidString` is for the additional app descriptors.

lib/application_group.ts
mobitar: Yup we are the same class that does the persisting.

PR #110, opened 2020-09-13
Session refactors

mobitar:
PR to go along with server-side changes: https://github.com/standardnotes/syncing-server/pull/107

Also binds all API responses to rigid types so that we're not accessing values blindly.


test/session.test.js
```diff
@@ -275,13 +275,15 @@ describe('server session', function() {
   });
 
   it('should fail if syncing while a session refresh is in progress', async function () {
-    this.application.apiService.refreshSession();
+    const refreshPromise = this.application.apiService.refreshSession();
     const syncResponse = await this.application.apiService.sync([]);
 
     expect(syncResponse.error).to.be.ok;
 
     const errorMessage = 'Your account session is being renewed with the server. Please try your request again.';
```
baptiste-grob, line 283: Is there no string constant we can access from the test files? Having a test break because of a wording change doesn't sound ideal.

baptiste-grob (commented):
@johnny243 Has worked on API matters much more than I have so I've only commented on superficial aspects of this PR and will defer to his judgement for the core contents 👍

lib/services/api/api_service.ts
```diff
@@ -128,12 +129,12 @@ export class SNApiService extends PureService {
...
         );
       });
-    return response;
+    return response as KeyParamsResponse;
   }
 
   async register(email: string, serverPassword: string, keyParams: SNRootKeyParams) {
     if (this.registering) {
-      return this.createErrorResponse(messages.API_MESSAGE_REGISTRATION_IN_PROGRESS);
+      return this.createErrorResponse(messages.API_MESSAGE_REGISTRATION_IN_PROGRESS) as RegistrationResponse;
```
baptiste-grob, line 137: This works because `RegistrationResponse`'s additional properties are all optional, however the day we stray from that unwritten rule we'll crash clients by feeding them false information. Perhaps a comment on `RegistrationResponse` could be enough but there should be something preventing `TypeError`s here.

lib/services/api/api_service.ts
```diff
@@ -160,7 +161,7 @@ export class SNApiService extends PureService {
     mfaCode?: string
   ) {
     if (this.authenticating) {
-      return this.createErrorResponse(messages.API_MESSAGE_LOGIN_IN_PROGRESS);
+      return this.createErrorResponse(messages.API_MESSAGE_LOGIN_IN_PROGRESS) as SignInResponse;
```
baptiste-grob, line 164: Same comment as with `RegistrationResponse`

test/session.test.js
mobitar: We just capture the promise value so we can await it at the end of the test.

lib/services/api/responses.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export type ChangePasswordResponse = RegistrationResponse;
+
+export type SignOutResponse = HttpResponse & {
+
+}
+
+export type SessionRenewalResponse = HttpResponse & {
+  session?: SessionBody
+}
```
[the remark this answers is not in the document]
mobitar: Does Git complain about that?

lib/services/api/api_service.ts
mobitar: For now it would seem that all response values will be optional if the type is &'ed with HttpResponse. This is because HttpResponse could always have an error, in which case all values would be undefined. Perhaps if we called it `RegistrationSuccessResponse`, we would then have to start thinking about non-optionality.

lib/services/api/session_manager.ts
```diff
@@ -94,7 +96,7 @@ export class SNSessionManager extends PureService {
 
   public async signOut() {
```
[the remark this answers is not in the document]
mobitar: Fixed by awaiting http request

lib/services/api/session.ts
```diff
@@ -1,44 +1,87 @@
...
-  public accessToken: string
-  public expireAt?: number
-  public refreshToken?: string
-  public validUntil?: number
+  /** Return the token that should be included in the header of authorized network requests */
+  public abstract get authorizationValue(): string;
 
-  static FromRaw(raw: any) {
-    return new Session(raw.accessToken, raw.expireAt, raw.refreshToken, raw.validUntil);
+  static FromRawStorageValue(raw: any) {
```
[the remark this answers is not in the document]
mobitar: Fixed here: https://github.com/standardnotes/snjs/pull/110/commits/92565a6f4f94a46918e2f9644b575993221ecb61

I had to go with `&` instead of `|` for combining the expected types, even though I would have preferred `|`. But the compiler didn't create they were all optional properties. Let me know if you know a different way.

lib/services/api/http_service.ts
```diff
@@ -118,7 +118,7 @@ export class SNHttpService extends PureService {
       resolve(response);
     } else {
       if (!response.error) {
-        response.error = { status: httpStatus };
+        response.error = { message: 'Unknown error', status: httpStatus };
```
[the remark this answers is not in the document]
mobitar: Fixed

johnny243 (approved):

lib/services/api/responses.ts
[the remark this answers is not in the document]
baptiste-grob: Yeah I think it messes up `git diff` https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file
So if you're a terminal git user like @karolsojko you'd be annoyed ;)

lib/services/api/session.ts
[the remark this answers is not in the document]
baptiste-grob: Do you mean that the compiler wouldn't let you check for things like `raw.jwt`? I think you should keep it as '|' because that's what callers expect, and if you can't find a way to make the compiler behave, force its hand. It's a small function and forcing input correctness with the compiler is more important IMO.
By the way, when working with `unknown` Typescript wants you to use the `in` operator https://www.typescriptlang.org/docs/handbook/advanced-types.html

test/session.test.js
baptiste-grob: I meant `const errorMessage`. But it looks like that belongs in a previous commit anyway.

lib/services/api/api_service.ts
baptiste-grob: That's fine but we should document it in the type declaration so we don't forget and add a non-optional property in the future.

PR #111, opened 2020-09-15
fix: send status changes in AccountSyncOperation

baptiste-grob:
AccountSyncOperation was only sending SyncSignal.Response and not SyncSignal.StatusChanged


lib/services/sync/account/operation.ts
```diff
@@ -57,6 +57,7 @@ export class AccountSyncOperation {
   }
 
   async run() : Promise<void> {
+    await this.receiver(SyncSignal.StatusChanged);
```
mobitar, line 60: Should this be right after popPayloads instead? Because we'd assume that when the status change event is received, the consumer would call our public functions to determine what's changed. Those would be get pendingUploadCount and totalUploadCount. But those two values will not have changed until you call popPayloads below. Do I have that right?

baptiste-grob: Imagine We have 160 payloads to upload (our current upload limit is 150.)
If we call the receiver after popPayloads, pendingUploadCount's value will be 10 (160 - 150), which isn't true because we still haven't uploaded those first 150 payloads. That's why I put it before.
And now that I think about it, assuming when the consumer would query Operation's public state is probably not the best idea. How about we instead make that state private and pass it to the receiver callback function?

mobitar: Hmm..yeah, seems to make sense. Up to you if you don't think that would be too disruptive a change. Not sure what it entails.

mobitar (approved):

PR #114, opened 2020-09-22
feature: hasPendingMigrations

radko93:
This PR enables the client to check if there are any migrations pending before starting them. This is only possible after `MigrationsLoaded` has happened.

### DIFF
--- a/lib/events.ts
+++ b/lib/events.ts
@@ -22,7 +22,7 @@ export enum ApplicationEvent {
   /**
    * The application has started `prepareForLaunch` and completed base migration.
    */
-  PreparingForLaunch = 23,
+  MigrationsLoaded = 23,
   /**
    * The applicaiton is fully unlocked and ready for i/o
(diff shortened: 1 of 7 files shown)


lib/events.ts
```diff
@@ -11,16 +11,20 @@ export enum ApplicationEvent {
...
    * Called when the application has initialized and is ready for launch, but before
    * the application has been unlocked, if applicable. Use this to do pre-launch
    * configuration, but do not attempt to access user data like notes or tags.
    */
   Started = 10,
 
-  /** 
-   * The applicaiton is fully unlocked and ready for i/o 
+  /**
+   * The application has started `prepareForLaunch` and completed base migration.
```
mobitar, line 23: This comment should be pertaining to migrations only:

```
The application has loaded all pending migrations (but not run any, except for the base one), and consumers may now call `hasPendingMigrations`
```

mobitar (approved):

PR #121, opened 2020-09-30
feat: duplicateNote

baptiste-grob:
Extracts the duplication logic from web (revision restoring) to snjs

(Still a draft while I finish writing tests)


baptiste-grob:
Didn't take into account itemManager's duplicateItem method. Will update

mobitar:
This is the function btw that handles retaining editor prefs: https://github.com/standardnotes/snjs/blob/004/lib/protocol/payloads/functions.ts#L19

PR #122, opened 2020-10-01
feat: duplicate item content overrides

baptiste-grob:
Expose ItemManager.duplicateItems to clients and let them specify content overrides (useful for restoring item revisions as copies or duplicating notes in the future)

Also adds test to ensure a duplicated note keeps its editors and tags.

### DIFF
--- a/lib/services/item_manager.ts
+++ b/lib/services/item_manager.ts
@@ -545,20 +545,25 @@ export class ItemManager extends PureService {
    * Duplicates an item and maps it, thus propagating the item to observers.
    * @param isConflict - Whether to mark the duplicate as a conflict of the original.
    */
-  public async duplicateItem(uuid: UuidString, isConflict = false) {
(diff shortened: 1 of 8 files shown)


lib/services/item_manager.ts
```diff
@@ -545,20 +545,24 @@ export class ItemManager extends PureService {
    * Duplicates an item and maps it, thus propagating the item to observers.
    * @param isConflict - Whether to mark the duplicate as a conflict of the original.
    */
-  public async duplicateItem(uuid: UuidString, isConflict = false) {
```
mobitar, line 548: So the reason we duplicate by uuid is that Item objects are immutable, and when a caller wants to duplicate an item, they typically mean "duplicate the latest values of this item", but the caller may be passing in a stale item reference. By looking up the item instead we ensure we are duplicating its latest value.

mobitar (approved):

PR #123, opened 2020-10-02
fix: duplicate revision decryption

mobitar:
- Fixes issue where revisions of an item that is created from a duplication event couldn't be decrypted

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -82,24 +82,25 @@ export { SNPrivilegesService } from './services/privileges_service';
 export { SNSingletonManager } from './services/singleton_manager';
 export { SNApiService } from './services/api/api_service';
 export {
+  addIfUnique,
+  arrayByDifference,
   Copy,
-  findInArray,
-  isNullOrUndefined,
+  dateSorted,
   deepMerge,
+  dictToArray,
   extendArray,
+  findInArray,
+  getGlobalScope,
+  greaterOfTwoDates,
+  isNullOrUndefined,
+  jsonParseEmbeddedKeys,
+  omitUndefinedCopy,
+  removeFromArray,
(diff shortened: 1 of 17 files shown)


lib/index.ts
```diff
@@ -82,24 +82,25 @@ export { SNPrivilegesService } from './services/privileges_service';
...
+  isNullOrUndefined,
   jsonParseEmbeddedKeys,
+  omitUndefinedCopy,
+  removeFromArray,
+  removeFromIndex,
+  subtractFromArray,
   topLevelCompare,
-  dateSorted
+  truncateHexString,
+  uniqCombineObjArrays,
```
mobitar, line 103: Just sorted this alphabetically..

lib/protocol/operator/004/operator_004.ts
```diff
@@ -196,13 +196,26 @@ export class SNProtocolOperator004 extends SNProtocolOperator003 {
   private async authenticatedDataToString(
     attachedData: ItemAuthenticatedData
   ) {
-    return this.crypto.base64Encode(JSON.stringify(sortedCopy(attachedData)));
+    return this.crypto.base64Encode(
+      JSON.stringify(
+        sortedCopy(
+          omitUndefinedCopy(
```
mobitar, line 202: Technically JSON.stringify will already omit any keys for which the value is undefined. But I didn't want to count on this being an implementation detail of web browsers. Perhaps React Native does it differently. So this is why we omit undefined keys explicitly.

johnny243 (changes requested):
Tests for remote history seems to be failing

mobitar:
@johnny243 you'd need to run tests in using development environment so that jobs can run, in order for revision tests to succeed.

PR #125, opened 2020-10-05
feature: mobile preferences migration

radko93:
As snjs has now aware of mobile preferences, is this a proper way to migrate old prefs? Tests have to be added of course.


lib/migrations/2020-01-15.ts
```diff
@@ -348,6 +353,29 @@ export class Migration20200115 extends Migration {
...
+    if (legacyOptionsState) {
+      const legacySortBy = legacyOptionsState.sortBy;
+      migratedOptionsState = {
+        sortBy: isNullOrUndefined(legacySortBy) || legacySortBy === 'updated_at' || legacySortBy === 'client_updated_at' ? CollectionSort.UpdatedAt : legacySortBy,
+        sortReverse: legacyOptionsState.sortReverse ?? false,
+        hideNotePreview: legacyOptionsState.hidePreviews ?? false,
+        hideDate: legacyOptionsState.hideDates ?? false
+      }
+    }
+    rawStructure.nonwrapped![StorageKey.MobilePreferences] = {
```
mobitar, line 373: Nonwrapped? Should be part of wrapped. Nonwrapped is only information needed pre-unlock. In this case all this info appears to be post-unlock relevant.

lib/storage_keys.ts
```diff
@@ -25,6 +25,7 @@ export enum StorageKey {
   BiometricsState = 'biometrics_state',
   MobilePasscodeTiming = 'passcode_timing',
   MobileBiometricsTiming = 'biometrics_timing',
+  MobilePreferences = 'preferencesKey',
```
mobitar, line 28: Unless you need it this way for a reason, the key value should just be `preferences` instead of `preferencesKey`

mobitar (approved):

PR #126, opened 2020-10-07
fix: custom filtering for notes when a tag is set

baptiste-grob:


baptiste-grob:
(I think I've figured out a better way to do this, removing request for review)

baptiste-grob:
The solution ends up being much more involved and I don't want to delay this fix. Will post a subsequent PR if it bears fruit

mobitar:
Can you elaborate on the ideal solution, for posterity?

baptiste-grob:
The ideal solution would be for `ItemCollectionNotesView` to construct a custom filter function that can handle tags in the most efficient manner possible, and pass it to `ItemCollection.setDisplayOptions(...)`.
Right now by filtering items after the fact, `ItemCollection` is wasting time sorting items that are bound to be removed later.

PR #128, opened 2020-10-08
fix: sort notes when displaying tag

baptiste-grob:


lib/protocol/collection/item_collection_notes_view.ts
```diff
@@ -78,13 +76,12 @@ export class ItemCollectionNotesView {
...
-          element.content_type === ContentType.Note &&
-          !element.deleted &&
-          !element.trashed &&
-          this.displayFilter(element)
-      ) as SNNote[];
+      const taggedNoteUuids = new Set(
+        this.collection.uuidReferencesForUuid(tag.uuid)
+      );
+      this.displayedList = notes.filter(
+        note => taggedNoteUuids.has(note.uuid)
```
mobitar, line 83: You could also just have used `tag.hasRelationshipWithItem(note)`

baptiste-grob: `tag.hasRelationshipWithItem(note)` will search its references list every time, so it's fine when you're checking a single note but probably not when you don't know how many notes you are checking

mobitar: It's filtering through an array either way no?

PR #129, opened 2020-10-09
fix: mobile jwt migration

mobitar:


radko93:
Migration of users with account + passcode fails when using this PR.
1. Create an account on the legacy app, add a passcode.
2. Upgrade to the new app.
3. Fill correct passcode

This line will fail https://github.com/standardnotes/snjs/blob/fix-jwt/lib/migrations/2020-01-15.ts#L351 with
>  uuid is null, yet this payloads fields indicate it shouldnt be.

radko93 (approved):

PR #130, opened 2020-10-12
Fix tags showing trashed notes

baptiste-grob:


lib/protocol/collection/item_collection_notes_view.ts
```diff
@@ -63,27 +82,25 @@ export class ItemCollectionNotesView {
...
-    } else if (tag) {
-      const taggedNoteUuids = new Set(
-        this.collection.uuidReferencesForUuid(tag.uuid)
-      );
+    if (!this.tagUuid) {
+      this.displayedList = notes;
+      return;
+    }
+
+    const tag = this.collection.find(this.tagUuid) as SNTag;
```
mobitar, line 96: I'm not sure that system smart tags (all, trash, archived) will appear inside the collection: https://github.com/standardnotes/snjs/blob/004/lib/services/item_manager.ts#L71

baptiste-grob: You're right, and this broke SNJS. I moved too fast. I pushed a subsequent hotfix afterwards though.

PR #132, opened 2020-10-15
feature: let apps set persistence policy to ephemeral

baptiste-grob:
Required for https://github.com/standardnotes/decrypt/pull/2

### DIFF
--- a/lib/application.ts
+++ b/lib/application.ts
@@ -1381,6 +1381,11 @@ export class SNApplication {
     return this.protocolService!.repersistAllItems();
   }
 
+  public enableEphemeralPersistencePolicy(): Promise<void> {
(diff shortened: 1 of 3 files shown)


lib/application.ts
```diff
@@ -1381,6 +1381,11 @@ export class SNApplication {
     return this.protocolService!.repersistAllItems();
   }
 
+  public setEphemeralPersistencePolicy(): Promise<void> {
```
mobitar, line 1384: Setters usually take an argument no? May make more sense to call this `setEphemeralPersistencePolicy`

mobitar (approved):

lib/application.ts
baptiste-grob: Oh yeah that's much better

PR #133, opened 2020-10-19
refactor: main auth completion handler should be responsible for setting root key

mobitar:
- Fixes issue where setting the root key on sign in/register would also call encryptAllItemsKeys (should only be called on password change)
- Fixes issue where performing a sign in through detached session recovery wouldn't set global root key. This is fixed via the fact that the main sign in function (through the auth completion handler) now sets the root key directly, as opposed to letting callers be responsible for this.

### DIFF
--- a/lib/services/api/session_manager.ts
+++ b/lib/services/api/session_manager.ts
@@ -206,6 +206,7 @@ export class SNSessionManager extends PureService<SessionEvent> {
         )
       };
     }
+    const wrappingKey = await this.challengeService.getWrappingKeyIfApplicable(true);
     email = cleanedEmailString(email);
     const rootKey = await this.protocolService!.createRootKey(
       email,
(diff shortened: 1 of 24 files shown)


lib/services/api/session_manager.ts
```diff
@@ -451,12 +445,17 @@ export class SNSessionManager extends PureService<SessionEvent> {
   }
 
   private async handleSuccessAuthResponse(
-    response: RegistrationResponse | SignInResponse | ChangePasswordResponse
+    response: RegistrationResponse | SignInResponse | ChangePasswordResponse,
+    rootKey: SNRootKey
   ) {
+    const wrappingKey = await this.challengeService.getWrappingKeyIfApplicable(true);
```
baptiste-grob, line 451: Currently the app asks for your passcode before it does anything related to signing in/registering. This has the beneficial side-effect of preventing users from launching a destructive operation without being able to complete it.

What happens if a user can't remember their passcode, or it they close the app without inputting their passcode here?

mobitar: Good point, I'll re-work this.

baptiste-grob (approved):

PR #141, opened 2020-10-28
refactor: backups functions

baptiste-grob:
Specify the backups functions for each of our use cases:
- `createEncryptedBackupString()` for web and mobile
- `createDecryptedBackupString()` for mobile
- `createDecryptedBackup()` for web, letting it use the items to generate a zip file

The protocol service backup function now encrypts items using Promise.all, which should improve speed on mobile where waiting on the RN bridge is costly

The decrypted backup functions ask for a passcode if any note is protected

- [x] Ask for biometrics as well (cc @radko93, do you see this being an issue in any way? For exporting a decrypted backup)


lib/application.ts
```diff
@@ -910,18 +910,43 @@ export class SNApplication {
...
+    if (backup) {
+      return JSON.stringify(backup, null, 2);
+    } else {
+      return undefined;
+    }
+  }
+
+  public async createDecryptedBackup(): Promise<BackupFile | undefined> {
+    if (
+      this.hasPasscode() &&
```
mobitar, line 938: If no passcode, we'd still want to prompt for account password if account.

lib/services/protocol_service.ts
```diff
@@ -727,50 +726,42 @@ export class SNProtocolService extends PureService implements EncryptionDelegate
...
+    intent: EncryptionIntent.FileDecrypted | EncryptionIntent.FileEncrypted
+  ): Promise<BackupFile> {
+    let items = this.itemManager!.items;
+    let keyParams: AnyKeyParamsContent | undefined;
+
+    if (intent === EncryptionIntent.FileDecrypted) {
+      items = items.filter(
+        (item) => item.content_type !== ContentType.ItemsKey
+      );
+      keyParams = (await this.getRootKeyParams())?.getPortableValue();
```
mobitar, line 739: If decrypted I suppose we wouldn't need keyParams in the file.

lib/application.ts
baptiste-grob: Do we want to prompt for both? If not then we need to require authentication for `removePasscode()` as well, otherwise it becomes easy to circumvent.

baptiste-grob: I decided to make it prompt for both for safety.

mobitar: Yeah, I guess it doesn't hurt. If you're bulk ejecting data we want to make it as safe as possible.

mobitar: @baptiste-grob the only downside is that some people download a decrypted backup when they have forgotten their account password..

radko93:
Biometrics don't actually encrypt anything at the moment, so no issue with encrypting backups.

baptiste-grob:
Closed in favor of https://github.com/standardnotes/snjs/pull/155

PR #145, opened 2020-11-03
feature: add error reporting singleton

baptiste-grob:
At first I wanted to throw or return errors from operators but it would have been an extremely disruptive change and have broken all the tests, so I went with a singleton where clients can set its error reporter like this:

```ts
SNErrors.onUnhandledError = (error) => {
  // Report error, e.g with Bugsnag
}
```

And SNJS can call like this, anywhere in the app:

```ts
SNErrors.onUnhandledError(Error('...'));
```

### DIFF
--- a/lib/protocol/operator/002/operator_002.ts
+++ b/lib/protocol/operator/002/operator_002.ts
@@ -18,6 +18,7 @@ import { PayloadFormat } from '@Payloads/formats';
 import { CreateEncryptionParameters, CopyEncryptionParameters } from '@Payloads/generator';
 import { ProtocolVersion } from '@Protocol/versions';
 import { SNRootKey } from '@Protocol/root_key';
+import { SNLog } from '@Lib/log';
 
 /**
  * @deprecated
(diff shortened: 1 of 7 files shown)


lib/protocol/operator/002/operator_002.ts
```diff
@@ -119,7 +120,9 @@ export class SNProtocolOperator002 extends SNProtocolOperator001 {
     }
     const localAuthHash = await this.crypto.hmac256(ciphertextToAuth, authKey);
     if (this.crypto.timingSafeEqual(authHash, localAuthHash!) === false) {
-      console.error('Auth hash does not match, returning null.');
+      const error = Error('Auth hash does not match.');
+      SNErrors.onUnhandledError(error);
+      console.error(error);
```
mobitar, line 125: Should we override console.error to call SNErrors?

radko93: Why we would want to do that? console.errors are only for development

mobitar (approved):

radko93 (approved):

PR #150, opened 2020-11-12
fix: prevent signing in twice

baptiste-grob:
Currently clients are able to call application.signIn() multiple times, even when they are already signed in. This causes problems like theme duplication and "Potentially Out of Sync" errors. 
@radko93 there were reports that looked like this on mobile, so you may want to check if there's any way this could happen.

(Still a draft because it's making some session tests fail)


mobitar:
Sign in should actually be idempotent. Because when your session expires (but still exists as a value locally), we have you sign in again.

So in fact we should design it such that if sign in were called 10 times consecutively, no harm is done.

baptiste-grob:
So I realized that SessionManager.signIn() is idempotent, it's SNApplication.signIn() that's not, as it calls downloadFirstSync. I could change the method to handle that case but since this is a client-facing method, from this perspective it should never be called twice in a row if the first signIn was successful, and so throwing an error from _that_ method would be appropriate. What do you think?

mobitar:
Hmm yeah that makes sense. You'd just have to be sure that any tests that were expecting idempotency with application method to be modified to use the sessionManager call directly.

lib/application.ts
```diff
@@ -1107,29 +1110,40 @@ export class SNApplication {
...
-      this.syncService!.resetSyncState();
-      await this.storageService!.setPersistencePolicy(
-        ephemeral
-          ? StoragePersistencePolicies.Ephemeral
-          : StoragePersistencePolicies.Default
-      );
-      if (mergeLocal) {
-        await this.syncService!.markAllItemsAsNeedingSync(true);
+    if (this.hasAccount()) {
+      throw Error('Tried to register when an account already exists.');
```
mobitar, line 1114: Should we use `throw SNLog.error(...)` here instead?

test/auth.test.js
```diff
@@ -81,6 +81,37 @@ describe('basic auth', () => {
     expect(await this.application.protocolService.getRootKey()).to.be.ok;
   }).timeout(20000);
 
+  it('cannot sign in twice to registered account', async function () {
```
mobitar, line 84: Can we also have a test that ensures that if a sign in fails (due to incorrect password), the `this.signingIn` mutex is correctly reset?

lib/application.ts
baptiste-grob: The way I see it `SNLog.error(...)` should be used for errors that are exceptional but do not warrant halting the current operation. Calling `application.signIn(...)` twice, on the other hand, is a logic fault that should be fixed by the client, hence the throw.
Any uncaught exception will be reported by our tools as well so this error wouldn't be lost in the ether if it happened on a user's device.

mobitar (approved):

PR #151, opened 2020-11-12
SNJS version

mobitar:
This PR:

- Moves from timestamp based migrations to SNJS version based migrations
- Adds automatic recovery if on launch we detect the keychain has been cleared for an account user (when user restores device from backup, like iCloud)


mobitar:
@baptiste-grob @radko93 new device interface method required to be implemented by clients:

```
  async legacy_setRawKeychainValue(value: any) {
    return Keychain.setKeys(value);
  }
```

lib/migrations/base.ts
```diff
@@ -0,0 +1,210 @@
...
+    const hasPasscode = await reader.hasPasscode();
+    if (hasPasscode) {
+      /** Doesn't apply if using passcode, as keychain would be bypassed in that case */
+      return;
+    }
+
+    const accountKeysMissing = !(await reader.hasNonWrappedAccountKeys());
+    if (!accountKeysMissing) {
+      return;
+    }
```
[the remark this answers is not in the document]
mobitar: I don't know, the idea is to make the variables describe the intention, in place of a comment. `accountKeysMissing` tells you exactly the problem, but `hasAccountKeys` doesn't tell you what the problem or premise is.

But it could easily go both ways I suppose.

mobitar:
Also please note that henceforth any SNJS changes should increment the package.json version number.

lib/migrations/base.ts
```diff
@@ -0,0 +1,244 @@
...
+            const allItems = (
+              await this.services.deviceInterface.getAllRawDatabasePayloads(this.services.identifier)
+            ) as any[];
+            let itemToDecrypt = allItems.find(item => {
+              const payload = CreateMaxPayloadFromAnyObject(item);
+              return ContentTypeUsesRootKeyEncryption(payload.content_type);
+            });
+            if (!itemToDecrypt) {
+              /** If no root key encrypted item, just choose any item */
+              itemToDecrypt = allItems[0];
```
baptiste-grob, line 203: How can this work? Or I suppose more importantly, when would this _not_ work?

lib/migrations/readers/functions.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export function CreateReader(
+  version: string,
+  deviceInterface: DeviceInterface,
+  identifier: ApplicationIdentifier,
+  environment: Environment
+): StorageReader {
+  const readerClass = ReaderClassForVersion(version);
+  return new readerClass(deviceInterface, identifier, environment);
+}
```
baptiste-grob, line 33: You might want to set `"files.insertFinalNewline": true` in your vscode config :)

lib/migrations/readers/index.ts
```diff
@@ -0,0 +1,2 @@
+export { StorageReader2_0_0 } from './reader_2_0_0';
+export { StorageReader1_0_0 } from './reader_1_0_0';
```
baptiste-grob, line 2: You might want to set `"files.insertFinalNewline": true` in your vscode config :)

lib/migrations/readers/reader.ts
```diff
@@ -0,0 +1,36 @@
...
+  constructor(
+    protected deviceInterface: DeviceInterface,
+    protected identifier: ApplicationIdentifier,
+    protected environment: Environment
+  ) {
+
+  }
+
+  public static version(): string {
+    throw 'Must override';
```
baptiste-grob, line 21: should be `throw Error('Must override')` instead

test/migrations/migration.test.js
```diff
@@ -6,34 +6,114 @@ const expect = chai.expect;
...
+    application.deinit();
+  });
+
+  it('after running all migrations from a 1.0.0 installation, should set stored version to current', async function () {
+    const application = await Factory.createAppWithRandNamespace();
+    /** Set up 1.0.0 structure with tell-tale storage key */
+    await application.deviceInterface.setRawStorageValue(
+      'migrations',
+      JSON.stringify(['anything'])
+    );
```
baptiste-grob, line 74: Does this cover for the installations with a migration timestamp?

dist/@types/device_interface.d.ts
```diff
@@ -50,6 +50,8 @@ export declare abstract class DeviceInterface {
     abstract removeRawDatabasePayloadWithId(id: string, identifier: ApplicationIdentifier): Promise<void>;
     abstract removeAllRawDatabasePayloads(identifier: ApplicationIdentifier): Promise<void>;
     abstract getNamespacedKeychainValue(identifier: ApplicationIdentifier): Promise<any>;
+    /** Allows SNJS to set the top level keychain value */
+    abstract legacy_setRawKeychainValue(value: any): Promise<void>;
```
baptiste-grob, line 54: I'm not sure I understand how to implement this on web. What version does legacy refer to? Should it be named something else, like `v1_0_0_setRawKeychainValue` instead?

lib/services/api/session_manager.ts
```diff
@@ -118,16 +118,13 @@ export class SNSessionManager extends PureService<SessionEvent> {
   }
 
   public getUser() {
-    if (this.user && !this.apiService.getSession()) {
-      this.alertService.alert(
-        ErrorAlertStrings.MissingSessionBody,
-        ErrorAlertStrings.MissingSessionTitle
-      );
-      SNLog.error(Error('User is defined but no session is present.'))
-    }
```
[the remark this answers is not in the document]
mobitar: We don't need this anymore as the new provisions in migrations should preclude this use case. This was also bad design because we're presenting an alert inside a getter.

lib/services/migration_service.ts
```diff
@@ -65,75 +83,58 @@ export class SNMigrationService extends PureService {
...
-      const aTimestamp = a.timestamp();
-      const bTimestamp = b.timestamp();
-      if (aTimestamp < bTimestamp) {
-        return -1;
-      } else if (aTimestamp > bTimestamp) {
-        return 1;
-      } else {
-        return 0;
-      }
+      return compareSemVersions(a.version(), b.version());
```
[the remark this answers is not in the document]
mobitar: What do you mean? Migrations are static files that are compiled. They've all been changed to be version-based.

dist/@types/device_interface.d.ts
mobitar: Yeah maybe it should be prefixed with version. Although I'm not sure if version makes sense because we're accessing this from the latest code. Legacy just means in this case "we don't do this anymore (but your code can if it really has to)". Maybe it should be called `unsafe_setRawKeychainValue` or something like that?

lib/migrations/base.ts
mobitar: The only time it shouldn't work is if you have 0 items. I suppose we're not handling that case here because I don't expect it, especially if you have an account. But we can put in safeguards.

lib/version.ts
```diff
@@ -0,0 +1,45 @@
...
+ */
+export function compareSemVersions(a: string, b: string) {
+  if (a === b) {
+    return 0;
+  }
+  if (isRightVersionGreaterThanLeft(a, b)) {
+    return -1;
+  } else {
+    return 1;
+  }
```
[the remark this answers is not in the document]
mobitar: I guess 2.0.01 is not really a semver in that case..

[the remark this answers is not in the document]
mobitar: Got a fix incoming nonetheless

test/migrations/migration.test.js
mobitar: This creates a legacy structure. "migrations" was a 1.0.0 key.

lib/migrations/readers/index.ts
mobitar: I actually don't like a final empty line 😳 Is this illegal?

baptiste-grob: It's mostly about compatibility with older unix tools I believe https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

lib/services/migration_service.ts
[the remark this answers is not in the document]
baptiste-grob: I mean in the user's storage. But by your response I assume there's a migration covering this case.

dist/@types/device_interface.d.ts
baptiste-grob: I think `legacy` is definitely a better candidate than `unsafe`, as the latter doesn't necessarily indicate that it has a deprecated/obsolete behavior.
The reason I'm suggesting a version number is, what happens if we change `setRawKeychainValue` again in the future?  A form of time-stamping prevents that question from being asked. But maybe we can cross that bridge when we get to it.

baptiste-grob (approved):

radko93 (approved):

PR #153, opened 2020-11-17
feat: add distributable package

karolsojko:
I removed the `dist` folder which will be generated with every build on `004` branch now and published as a package on Github Packages.

Added the `nvmrc` setup with a lock on latest stable `12.19.0` for node

once this is merged `snjs` will be possible to use an npm package like this:
```
// package.json
...
"@standardnotes/snjs": "^2.0.4",
...
```


baptiste-grob (changes requested):
SNJS should be install-able without needing to create a GitHub token, otherwise anyone wanting to contribute to downstream clients (web/mobile/desktop) will have to jump through an extra hoop to run/edit the projects

package.json
```diff
@@ -1,8 +1,14 @@
 {
-  "name": "snjs",
-  "version": "2.0.3",
+  "name": "@standardnotes/snjs",
+  "version": "2.0.4",
+  "engines": {
```
baptiste-grob, line 4: What does the engines key do?

karolsojko: https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines ;)

karolsojko:
> SNJS should be install-able without needing to create a GitHub token, otherwise anyone wanting to contribute to downstream clients (web/mobile/desktop) will have to jump through an extra hoop to run/edit the projects

Will ammend with creating an npm registry package instead of a Github package

karolsojko:
@baptiste-grob the package will be now published to NPM registry

package.json
baptiste-grob: Hah, fair enough, sorry~ in that case I assume this is strictly used to help one build with a correct version of node, in which case maybe `>=12.19.0 <14.x` would likely be more correct? I'm running node 14.14.0 and building without issue for example

baptiste-grob (approved):
Awesome 👍

PR #154, opened 2020-11-17
feat: add eslint checks

karolsojko:
I noticed there is no code style checks to prevent the code from going messy and even error prone, so I added:
- linter setup
- pre-push hook to run the linter
- run linter on builds
- fixed linter errors (left all the warnings for us to gradually improve)

### DIFF
--- a/lib/application.ts
+++ b/lib/application.ts
@@ -285,11 +285,11 @@ export class SNApplication {
   }
 
   public onStart() {
-    console.log('onStart not implemented');
+    console.error('onStart not implemented');
   }
 
   public onLaunch() {
-    console.log('onLaunch not implemented');
+    console.error('onLaunch not implemented');
   }
 
   private async handleLaunchChallengeResponse(response: ChallengeResponse) {
--- a/lib/application.ts
+++ b/lib/application.ts
@@ -285,11 +285,11 @@ export class SNApplication {
   }
 
   public onStart() {
-
+    console.log('onStart not implemented');
   }
 
   public onLaunch() {
-
+    console.log('onLaunch not implemented');
   }
 
   private async handleLaunchChallengeResponse(response: ChallengeResponse) {
--- a/lib/application.ts
+++ b/lib/application.ts
@@ -285,11 +285,11 @@ export class SNApplication {
   }
 
   public onStart() {
-    console.error('onStart not implemented');
+    // optional override
   }
 
   public onLaunch() {
-    console.error('onLaunch not implemented');
+    // optional override
   }
 
   private async handleLaunchChallengeResponse(response: ChallengeResponse) {
--- a/lib/services/history/history_manager.ts
+++ b/lib/services/history/history_manager.ts
@@ -147,12 +147,12 @@ export class SNHistoryManager extends PureService {
     if (entry && this.persistable) {
       /** Debounce, clear existing timeout */
       if (this.saveTimeout) {
-        if (this.timeout.hasOwnProperty('cancel')) {
+        if (Object.prototype.hasOwnProperty.call(this.timeout, 'cancel')) {
           this.timeout.cancel(this.saveTimeout);
         } else {
           clearTimeout(this.saveTimeout);
         }
-      };
+      }
       this.saveTimeout = this.timeout(() => {
         this.saveToDisk();
       }, PERSIST_TIMEOUT);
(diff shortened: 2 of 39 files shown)


lib/application.ts
```diff
@@ -285,11 +285,11 @@ export class SNApplication {
   }
 
   public onStart() {
-
+    console.log('onStart not implemented');
```
radko93, line 288: Maybe console error instead?

radko93 (commented):
I liked linters in general so it's a good change IMO

lib/application.ts
```diff
@@ -285,11 +285,11 @@ export class SNApplication {
...
 
   public onStart() {
-
+    console.error('onStart not implemented');
   }
 
   public onLaunch() {
-
+    console.error('onLaunch not implemented');
   }
```
baptiste-grob, line 293: Clients don't have to implement onStart or onLaunch for SNJS to work properly. Why log it?

lib/services/history/history_manager.ts
```diff
@@ -147,12 +147,12 @@ export class SNHistoryManager extends PureService {
     if (entry && this.persistable) {
       /** Debounce, clear existing timeout */
       if (this.saveTimeout) {
-        if (this.timeout.hasOwnProperty('cancel')) {
+        if (Object.prototype.hasOwnProperty.call(this.timeout, 'cancel')) {
```
baptiste-grob, line 150: I assume a linter error triggered this change? This is much less straightforward to read, there has to be a way to keep the existing syntax, maybe in the type signature?

lib/services/pure_service.ts
```diff
@@ -60,7 +60,7 @@ export abstract class PureService<E = string> {
   * at a certain migratory state.
   */
   public async handleApplicationStage(_stage: ApplicationStage) {
-
+    console.log('Function handleApplicationStage not implemented');
   }
```
baptiste-grob, line 64: Services don't have to implement this method so this could create unnecessary spam in the logs

baptiste-grob (commented):
Nice 👍

package.json
```diff
@@ -8,7 +8,8 @@
     "bundle": "webpack --config webpack.prod.js && npm run tsc",
     "docs": "jsdoc -c jsdoc.json",
     "tsc": "tsc --project lib/tsconfig.json && tscpaths -p lib/tsconfig.json -s lib -o dist/@types",
-    "test": "node test-server.js"
+    "test": "node test-server.js",
+    "lint": "eslint . --ext .ts"
```
johnny243, line 12: We can just have `eslint` here and in `.eslintrc` set the `overrides` option like so:

```json
...
"overrides": [
    {
      "files": ["lib/**/*.ts"]
    }
]
...
```

This will lint `*.ts` files inside the `lib` directory only.

lib/application.ts
karolsojko: Because of this error: https://github.com/typescript-eslint/typescript-eslint/blob/v3.10.1/packages/eslint-plugin/docs/rules/no-empty-function.md

I'm open to suggestion on how to change this

lib/services/history/history_manager.ts
karolsojko: Yes, this was caused by this: https://eslint.org/docs/rules/no-prototype-builtins

used the solution suggested in examples. I'm open to suggestions on how to change this

lib/services/pure_service.ts
karolsojko: Same as other empty functions: https://github.com/typescript-eslint/typescript-eslint/blob/v3.10.1/packages/eslint-plugin/docs/rules/no-empty-function.md

I'm open to suggestion on how to change this

package.json
karolsojko: @johnny243 have you tried this option?

I've tried `eslint .` and `eslint lib` with: 
- `"files": ["lib/**/*.ts"]`
- `"files": ["**/*.ts"]`
- `"files": ["*.ts"]`

none of those work as expected (pick up only js files in the root)

radko93 (approved):

lib/services/pure_service.ts
baptiste-grob: In this case I would silence the warning with a comment explaining that it's an optional override 👍

lib/application.ts
baptiste-grob: I would silence the warning with a comment explaining that it's an optional override

lib/services/history/history_manager.ts
baptiste-grob: I guess the underlying issue is using `hasOwnProperty`. And checking the prototype chain is something we should do in this case, so what do you think of this:
```js
if ('cancel' in this.timeout) {
  // ...
}
```

karolsojko:
@baptiste-grob applied your suggestions

baptiste-grob (approved):

PR #155, opened 2020-11-17
fix: correct backup encryption intent, better performance

baptiste-grob:
- When a backup file is requested from a client with no account or passcode, create decrypted backup file instead of encrypted backup file
- Better export performance: encrypt items using Promise.all to maximize throughput (cc @radko93), do the minimal amount of work while processing notes in the loop
- Adds more tests to ensure created backup files are correct


baptiste-grob:
Heads up @radko93, this changes the method signature for `createBackupFile()`. It always returns an object now and you can decide how/if you want to JSON stringify it, add properties to it, etc.

karolsojko:
Please rebase this PR onto to the current default branch which is `master`. Sorry for the trouble.

PR #159, opened 2020-11-23
fix: update node version

karolsojko:


radko93:
What if Mo or someone else has node 12? Not sure why we have to be so constrained if there are no compatibility issues.

karolsojko:
> What if Mo or someone else has node 12? Not sure why we have to be so constrained if there are no compatibility issues.

There will definitely be differences when it comes to installing some of the dependency libraries. So I would suggest we all agree on one version. For example I've seen issues that some libraries would work on one version of node and not the other. Also package-lock.json might look different when dependencies are installed on different versions of node so that might also cause confusions for a contributor.

It shouldn't be an issue to lock upon specific version range of Node if we are using nvm (see `.nvmrc` file) - it's just a matter of deciding which version to use.

karolsojko:
https://itnext.io/node-engines-helping-developers-everywhere-avoid-phantom-bugs-2eef519604b2

baptiste-grob (approved):
I agree that we should settle on one node version, and since 14 is the latest one I'm not opposed to using that

mobitar:
I'm ok with using the latest. Actually I just installed Node 15 the other day but maybe that was too progressive as I was having trouble building the desktop app with that 😅

PR #163, opened 2020-11-24
fix: safe use of postMessage

radko93:
There are some random cases when `component.window` is undefined and accessing `postMessage` fails (even if there is a check up there). This can happen when app gets locked suddently or when the users leaves the screen before an event is sent from the editor.


mobitar (approved):

baptiste-grob (approved):
Should we log and error when window is undefined?

PR #170, opened 2020-12-08
feat: synced preferences

baptiste-grob:
Extracts web preferences to SNJS as a generalized synced preferences mechanism.

Public API is `SNApplication.getPreference()/setPreference()`.

### DIFF
--- a/lib/services/preferences_service.ts
+++ b/lib/services/preferences_service.ts
@@ -88,8 +88,7 @@ export class SNPreferencesService extends PureService<'preferencesChanged'> {
       );
       if (
         previousRef?.uuid !== this.preferences.uuid ||
-        this.preferences.lastSyncBegan?.getTime() !==
-          previousRef?.lastSyncBegan?.getTime()
+        this.preferences.userModifiedDate > previousRef.userModifiedDate
       ) {
         this.notifyEvent('preferencesChanged');
       }
--- a/lib/services/preferences_service.ts
+++ b/lib/services/preferences_service.ts
@@ -63,12 +63,13 @@ export class SNPreferencesService extends PureService<'preferencesChanged'> {
     value: PrefValue[K]
   ): Promise<void> {
     if (!this.preferences) return;
-    await this.itemManager.changeItem<UserPrefsMutator>(
+    this.preferences = (await this.itemManager.changeItem<UserPrefsMutator>(
       this.preferences.uuid,
       (m) => {
         m.setPref(key, value);
       }
-    );
+    )) as SNUserPrefs;
+    this.notifyEvent('preferencesChanged');
   }
 
   private async reload() {
--- a/lib/services/preferences_service.ts
+++ b/lib/services/preferences_service.ts
@@ -50,10 +50,7 @@ export class SNPreferencesService extends PureService<'preferencesChanged'> {
     key: K,
     defaultValue: PrefValue[K] | undefined
   ): PrefValue[K] | undefined;
-  getValue<K extends PrefKey>(
-    key: K,
-    defaultValue: PrefValue[K]
-  ): PrefValue[K];
+  getValue<K extends PrefKey>(key: K, defaultValue: PrefValue[K]): PrefValue[K];
   getValue<K extends PrefKey>(
     key: K,
     defaultValue?: PrefValue[K]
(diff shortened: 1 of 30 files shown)


lib/services/preferences_service.ts
```diff
@@ -0,0 +1,100 @@
...
+    value: PrefValue[K]
+  ): Promise<void> {
+    if (!this.preferences) return;
+    this.preferences = (await this.itemManager.changeItem<UserPrefsMutator>(
+      this.preferences.uuid,
+      (m) => {
+        m.setPref(key, value);
+      }
+    )) as SNUserPrefs;
+    this.notifyEvent('preferencesChanged');
```
mobitar, line 72: Is the string value of this enforced somehow? Would TS complain if you misspelled this?

lib/services/singleton_manager.ts
```diff
@@ -186,73 +187,68 @@ export class SNSingletonManager extends PureService {
...
     createContentType: ContentType,
     createContent: PayloadContent
-  ) {
-    // eslint-disable-next-line no-async-promise-executor
-    return new Promise(async (resolve) => {
-      const matchingItems = this.validItemsMatchingPredicate(predicate);
-      if (matchingItems.length > 0) {
-        return resolve(matchingItems[0]);
-      }
-      /** Item not found, safe to create after full sync has completed */
```
mobitar, line 203: This comment didn't make it in the new code, I think we want to retain it. This logic can be intricate so we want to make sure we have as much context as possible.

test/app-group.test.js
```diff
@@ -88,6 +88,7 @@ describe('application group', () => {
     const identifier = application.identifier;
     await application.deinit(DeinitSource.SignOut);
 
+    await Factory.sleep(0.0001);
```
mobitar, line 91: Context? Sleep calls in tests should either not be used, be self-evident, or have a comment describing why sleeping is important. Especially if this test passed before with no sleep.

test/model_tests/appmodels.test.js
```diff
@@ -6,7 +6,7 @@ chai.use(chaiAsPromised);
 const expect = chai.expect;
 
 describe('app models', () => {
-  const BASE_ITEM_COUNT = 1; /** Default items key */
+  const BASE_ITEM_COUNT = 2; /** Default items key, user preferences */
```
mobitar, line 9: I don't remember where, but I started doing this in the style of

```
 const BASE_ITEM_COUNT = ['default-items-key', 'user-preferences'].length;
```

instead to make it self-documenting. Not saying this is necessary here but if we feel the need to refactor this that might be a good way to go.

test/preferences.test.js
```diff
@@ -0,0 +1,95 @@
...
+    callTimes += 1;
+    await Factory.sleep(0); /** Await next tick */
+    expect(callTimes).to.equal(1); /** App start */
+    await register.call(this);
+    expect(callTimes).to.equal(2);
+    await this.application.setPreference('editorLeft', 300);
+    expect(callTimes).to.equal(3);
+  });
+
+  it('discards existing preferences when logging in', async function () {
```
mobitar, line 82: Should we also have a test for "retains existing preferences when registering"?

test/singletons.test.js
```diff
@@ -309,6 +309,7 @@ describe('singletons', function() {
       }
     );
     await this.application.modelManager.emitPayload(notErrored);
+    await Factory.sleep(0);
```
mobitar, line 312: Needs context comment. Is it perhaps to allow decryptErroredItems to finish?

test/sync_tests/online.test.js
```diff
@@ -251,7 +251,7 @@ describe('online syncing', function() {
     // set item to be merged for when sign in occurs
     await this.application.syncService.markAllItemsAsNeedingSync();
     expect(this.application.syncService.isOutOfSync()).to.equal(false);
-    expect(this.application.itemManager.getDirtyItems().length).to.equal(2);
+    expect(this.application.itemManager.getDirtyItems().length).to.equal(3);
```
mobitar, line 254: Maybe can be `BASE_ITEM_COUNT + 1`?

mobitar (approved):

lib/services/preferences_service.ts
baptiste-grob:
```ts
class SNPreferencesService extends PureService<'preferencesChanged'>
```
Is what will make TS compain. This is probably not super straightforward but in typescript a string literal can be used as a type. so notifyEvent can only emit events of type `'preferencesChanged'`

test/preferences.test.js
baptiste-grob: We don't merge preferences at the moment, account preferences are always prioritized.

mobitar: Preferences would be auto-merged into registered accounts via the "Merge local items" option on sign in.

PR #178, opened 2020-12-17
feat: improve backups API

baptiste-grob:
Redo of https://github.com/standardnotes/snjs/pull/155


lib/application.ts
```diff
@@ -930,19 +930,26 @@ export class SNApplication {
...
+    if (intent === EncryptionIntent.FileDecrypted) {
+      if (items.some(item => item.protected) && this.hasPasscode()) {
+        const passcode = this.challengeService.promptForPasscode(
+          ChallengeReason.CreateDecryptedBackupWithProtectedItems
+        );
+        if (!passcode) {
+          return;
+        }
+      }
+      items = items.filter(item => item.content_type !== ContentType.ItemsKey);
```
mobitar, line 949: Hmm, this is too smart for the application instance to know. Largely the application instance should just be a delegator. Why not have the protocol service do this logic?

baptiste-grob: I wanted the protocol service to not have to inspect its parameters and be more straightforward but upon reflection this isn't the right compromise either yes

lib/application.ts
```diff
@@ -930,19 +930,25 @@ export class SNApplication {
...
-      intent,
-      returnIfEmpty
-    );
+    intent: EncryptionIntent
+  ): Promise<BackupFile | undefined> {
+    const items = this.itemManager.items;
+
+    if (intent === EncryptionIntent.FileDecrypted) {
+      if (items.some(item => item.protected) && this.hasPasscode()) {
+        const passcode = this.challengeService.promptForPasscode(
```
mobitar, line 942: Shouldn't this be awaited? And how is passcode being used? The return type for promptForPasscode isn't even a passcode, it's an object containing the passcode. Sounds like this functionality should be covered by a test.

lib/application.ts
```diff
@@ -930,19 +930,25 @@ export class SNApplication {
...
+          ChallengeReason.CreateDecryptedBackupWithProtectedItems
+        );
+        if (!passcode) {
+          return;
+        }
+      }
+    }
+
+    return this.protocolService.createBackupFile(intent, items);
```
mobitar, line 951: Doesn't seem like we need to pass items here since they're unfiltered.

lib/application.ts
mobitar: Oh I see, promptForPasscode is refactored in this PR. But still needs to be awaited.

baptiste-grob:
Like I said in the previous PR:

> Heads up @radko93, this changes the method signature for createBackupFile(). It always returns an object now and you can decide how/if you want to JSON stringify it, add properties to it, etc.

lib/application.ts
```diff
@@ -1193,7 +1203,7 @@ export class SNApplication {
         if (mergeLocal) {
           await this.syncService.markAllItemsAsNeedingSync(true);
         } else {
-          this.itemManager.removeAllItemsFromMemory();
+          void this.itemManager.removeAllItemsFromMemory();
```
mobitar, line 1206: Explain?

lib/services/protocol_service.ts
```diff
@@ -736,41 +736,42 @@ export class SNProtocolService extends PureService implements EncryptionDelegate
...
+        }
+      })
+    );
+
+    const keyParams = await this.getRootKeyParams();
+
     const data: BackupFile = {
       version: this.getLatestVersion(),
-      items: encryptedPayloads.map((p) => p.ejected())
+      items: await ejectedPayloads,
```
mobitar, line 769: Why not `await` at the definition level instead? In this case the `ejectedPayloads` variable is misnamed, as it's a promise, not ejectedVariables. Seems like the await should be at line 747.

I don't think there's any performance optimization to be gained here either from postponing the await. JS is ultimately single-threaded.

test/backups.test.js
```diff
@@ -25,29 +25,104 @@ describe('backups', () => {
     this.application = null;
   });
 
-  const BASE_ITEM_COUNT = 4;
+  const BASE_ITEM_COUNT_ENCRYPTED = [
+    'Note',
```
mobitar, line 29: This is kind of weird. Here you're limiting all tests in this file to only deal with 2 notes. The idea with `BASE` is that these are the number of items that the system is creating that your test doesn't know about, so if you're counting items, make sure you account for these.

mobitar: I don't think tests in this page should be limited to creating only 2 notes, and should be able to choose at their discretion how many notes to create. If they choose to create 10 notes, they can, and they should be able to ascertain that there are 10 + BASE items total.

lib/application.ts
baptiste-grob: `@typescript-eslint/no-floating-promises` will fire a warning when a promise is created but not awaited on or returned. This ensures that all async operations are accounted for.

In the case where we don't care to wait for the promise to return (fire-and-forget scenario), using the `void` operator (which does nothing) is used to silence the warning by explicitly communicating that we don't care about when the promise resolves or what its result is.

lib/services/protocol_service.ts
baptiste-grob: @radko93 has reported notable performance improvements from "parallel" processing. 

While JS is single-threaded, when an async function is awaiting on a disk, network, or encryption result (like when you call into the RN bridge), it's put in a queue and other JS sync code can be run in the meantime. This is why using things like Promise.all, while not fully parallelizing the work, schedules it smarter, resulting in better performance. Does that make sense?

All that being said, in this instance `getRootKeyParams` doesn't seem to be a heavy operation, it just comes at the tail-end of a very heavy one (depending on how many notes/tags you have), so I went with as much smart scheduling as possible. If you think that makes it hard to read let me know and I'll `await` at the definition instead like you proposed.

test/backups.test.js
baptiste-grob: You're right, this is making too many assumptions about the tests in this file

baptiste-grob: Are you saying the distinction between encrypted and decrypted base item count is weird?

mobitar: > Are you also saying the distinction between encrypted and decrypted base item count is weird?

No, wasn't looking at that

mobitar (approved):

baptiste-grob:
Watch out @radko93, `application.createBackupFile()` is slightly different as of 2.0.37 now :)

PR #181, opened 2020-12-21
feat: environmentFromString function

johnny243:
- adds a new function `environmentFromString`
- fixes type `(map as any)` on `platformFromString`

At the moment, `environmentFromString` will be used [here](https://github.com/johnny243/sn-components-api/blob/68b9dd1e5ea89ae99bb022a6f828da3883a6a87b/test/manager.test.ts#L140).

Also, `environmentFromString` is added to complement `environmentToString`.

### DIFF
--- a/lib/platforms.ts
+++ b/lib/platforms.ts
@@ -16,7 +16,7 @@ export enum Platform {
 }
 
 export function platformFromString(string: string) {
-  const map: Record<string, number> = {
+  const map: Record<string, Platform> = {
     'mac-web': Platform.MacWeb,
     'mac-desktop': Platform.MacDesktop,
     'linux-web': Platform.LinuxWeb,
@@ -44,7 +44,7 @@ export function platformToString(platform: Platform) {
 }
 
 export function environmentFromString(string: string) {
-  const map: Record<string, number> = {
+  const map: Record<string, Environment> = {
     'web': Environment.Web,
     'desktop': Environment.Desktop,
     'mobile': Environment.Mobile,
--- a/lib/platforms.ts
+++ b/lib/platforms.ts
@@ -43,6 +43,15 @@ export function platformToString(platform: Platform) {
   return map[platform];
 }
 
+export function environmentFromString(string: string) {
+  const map = {
+    'web': Environment.Web,
+    'desktop': Environment.Desktop,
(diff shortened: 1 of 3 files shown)


lib/platforms.ts
```diff
@@ -43,6 +43,15 @@ export function platformToString(platform: Platform) {
   return map[platform];
 }
 
+export function environmentFromString(string: string) {
+  const map = {
+    'web': Environment.Web,
+    'desktop': Environment.Desktop,
+    'mobile': Environment.Mobile,
+  };
+  return (map as any)[string];
```
mobitar, line 52: Why the `as any`? Should be a way to do it without?

johnny243: Without it, an error is shown:

`Element implicitly has an 'any' type because expression of type 'string' can't be used to index type...` and `No index signature with a parameter of type 'string' was found on type...`

---

We can add a type to `map` to get rid of the error, like so:

```javascript
const map: Record<string, number> = {
    'web': Environment.Web,
    'desktop': Environment.Desktop,
    'mobile': Environment.Mobile,
};
return map[string];
```

Also, we'll need to update `platformFromString` function to remove the `(map as any)`.

mobitar: > We can add a type to map to get rid of the error

Yes, that's the correct solution.

You should also add context to this PR so that in the future we can understand why we added this.

johnny243: Updated 👍🏽

lib/platforms.ts
```diff
@@ -43,6 +43,15 @@ export function platformToString(platform: Platform) {
   return map[platform];
 }
 
+export function environmentFromString(string: string) {
+  const map: Record<string, number> = {
```
mobitar, line 47: This should actually be `Record<string, Environment>` if that works.

lib/platforms.ts
```diff
@@ -16,7 +16,7 @@ export enum Platform {
 }
 
 export function platformFromString(string: string) {
-  const map = {
+  const map: Record<string, number> = {
```
mobitar, line 19: Should be Record<string, Platform>

mobitar:
package.json version number should also be incremeted

mobitar (approved):

PR #182, opened 2020-12-21
feat: reroute sign in on js test suite 

karolsojko:


mobitar (approved):
Does version need to be incremented for meta-level changes btw?

karolsojko:
No, but the publish step will fail otherwise

PR #185, opened 2020-12-28
feat: start ephemeral sessions with the server

baptiste-grob:


lib/services/api/session_manager.ts
```diff
@@ -166,7 +166,13 @@ export class SNSessionManager extends PureService<SessionEvent> {
             const email = challengeResponse.values[0].value as string;
             const password = challengeResponse.values[1].value as string;
             const currentKeyParams = await this.protocolService.getAccountKeyParams();
-            const signInResult = await this.signIn(email, password, false, currentKeyParams?.version);
+            const signInResult = await this.signIn(
+              email,
+              password,
+              false,
+              false,
```
mobitar, line 173: Hmm, so this is tricky. The function `reauthenticateInvalidSession` is called when the server reports the token is invalid. In the case of an ephemeral session, the server will probably report the token is invalid, then the client will display this. The user will likely enter their credentials, and now, their session is no longer ephemeral.

A solution here could be to determine whether our last valid session was ephemeral, and to use that value here instead of false.

baptiste-grob: Should the sync server be the one to report if a session is ephemeral then? And that's what SNJS would refer to in this case.

karolsojko: I can return a property `ephemeral: true` on the session object if you'd like ?

baptiste-grob: Per our private discussion I've gone with the "client remembers if session is ephemeral or not" approach instead

lib/services/api/session_manager.ts
```diff
@@ -67,6 +67,7 @@ export class SNSessionManager extends PureService<SessionEvent> {
 
   private user?: User
   private isSessionRenewChallengePresented = false;
+  private currentSessionIsEphemeral = false;
```
mobitar, line 70: Eh, we have better ways of detecting this imo.

- https://github.com/standardnotes/snjs/blob/master/lib/services/storage_service.ts#L131
- Otherwise it should be added as a property of the session object itself.

baptiste-grob: I ruled out adding ephemeral as a property on the session object because it made for something pretty messy, but the storage service method is exactly what I wanted, thanks :)

mobitar (approved):

PR #186, opened 2020-12-30
fix: present confirm dialog before revoking session

baptiste-grob:
(Nevermind the branch name, I got my features mixed up)

### DIFF
--- a/lib/application.ts
+++ b/lib/application.ts
@@ -489,28 +489,10 @@ export class SNApplication {
     return this.sessionManager.getSessionsList();
   }
 
-  /**
-   * Prompts the user to confirm their intent, and then revokes the session.
-   * @param sessionId The session's ID.
-   * @returns a response promise if the user confirmed their intent, undefined
-   * otherwise.
-   */
   public async revokeSession(
     sessionId: UuidString,
-  ): Promise<{
-    response?: Promise<HttpResponse>
-  }> {
-    const revoke = await this.alertService.confirm(
-      SessionStrings.RevokeText,
-      SessionStrings.RevokeTitle,
-      SessionStrings.RevokeConfirmButton,
-      ButtonType.Danger,
-    );
-    return {
-      response: revoke
-        ? this.sessionManager.revokeSession(sessionId)
-        : undefined,
-    };
+  ): Promise<HttpResponse> {
+    return this.sessionManager.revokeSession(sessionId);
   }
 
   public async userCanManageSessions(): Promise<boolean> {
--- a/lib/application.ts
+++ b/lib/application.ts
@@ -489,18 +489,28 @@ export class SNApplication {
     return this.sessionManager.getSessionsList();
   }
 
+  /**
+   * Prompts the user to confirm their intent, and then revokes the session.
+   * @param sessionId The session's ID.
+   * @returns a response promise if the user confirmed their intent, undefined
+   * otherwise.
+   */
   public async revokeSession(
-    sessionId: UuidString
-  ): Promise<HttpResponse | undefined> {
+    sessionId: UuidString,
+  ): Promise<{
+    response?: Promise<HttpResponse>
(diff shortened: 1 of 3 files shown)


baptiste-grob:
Here's how this API works in practice:
```ts
const result = await application.revokeSession(sessionId);
if (result.response) {
  /** User confirmed the prompt, optimistically update the UI */
  const response = await result.response;
  if ('error' in response) {
    /** Undo optimistic UI update and display error message */
  }
}
```
@radko93 I'd really like to get your feedback on this API. My first attempt had a mixture of callbacks and promises and that didn't feel right.

mobitar:
API looks a bit awkward to me. Awaiting `revokeSession` should be all you should await. Don't see why we're awaiting result.response again.

Should be something like

```javascript
const result = await application.revokeSession(sessionId);
if (result.response) {
  if (result.response.error) {

  } else {
  
  }
}
```

baptiste-grob:
This is simpler (and what I thought we needed at first) but very limiting in terms of UI. In my case, after the user has confirmed that they want to revoke their session, I want to remove the session from the UI list right away. Another use case would be showing a spinner, for example. You can't do these things before the user has said yes on the confirm prompt, and that happens inside the revokeSession method, hence the need for a two-stage API imo.

mobitar:
Is removing from list even before server confirmation the right UX? Users could close out the window thinking their session was revoked but was not, especially if dealing with high latency. Not saying it's an easy problem nonetheless. One other approach might be for revokeSession to accept a block `onConfirmed` that is called that allows the client to update the UI while the network request is in sync.

baptiste-grob:
Maybe some kind of loading indicator would be better yes.

Re the `onConfirmed` block, it would make the API look like this:

```ts
const result = await application.revokeSession(sessionId, () => {
  // UI update 1: spinner, list removal, etc.
});

if (result.response) {
  if (result.response.error) {
    // UI update 2a: restore list
  } else {
    // UI update 2b: update list
  }
}
```

What made me decide against it is that this API has a hidden, pernicious assumption: that _UI update 1_ will finish before _UI update 2_ starts. 
If for some reason _UI update 1_ doesn't finish before the request does (imagine a removal animation that takes 300ms to complete for example), then _UI update 2_ will happen too soon and potentially create undefined visual behavior.

Alternatively, we could have `onConfirmed` return a promise that SNJS will await before returning the network request. This is better, but there is still the assumption that clients can encapsulate their UI work in a single promise, which is too limiting considering how intricate UI changes can be. In our concrete cases (web/mobile) it's impossible as we don't have control over how/when React updates the UI.

I could definitely see the argument be made that `onConfirmed` returning a promise is sufficient, but I feel the two-stages, two promises solution is still the better approach that does not meddle with the client's specific UI updating concerns. And on the subject of awkwardness, I don't think the callback looks any cleaner/less awkward on the calling site either.

What if the API looked like this instead:
```ts
const request = await application.revokeSession(sessionId);
if (request.approved) {
  // UI update 1
  const response = await request.response;
  if ('error' in response) {
    // UI update 2a
  } else {
    // UI update 2b
  }
}
```

radko93 (commented):
After reviewing this API and reading your comments seems like keeping the confirmation on the client side was not that bad. We could also divide the alert and the revoke method, seems like combining it makes it very confusing for everyone.

mobitar:
In that case either we have the strings come from SNJS or SNJS has a `getDeleteSessionConfirmation` method. Just don't like the idea of duplicating long UI strings across codebases.

baptiste-grob:
Very well, See updated files where I added new strings to `SessionsStrings`, which is already an SNJS export

radko93 (approved):

radko93:
Please merge asap.

baptiste-grob:
(Can confirm that the tests pass on my local machine with `yarn test`.)

PR #187, opened 2020-12-30
chore: update readme links

funktionzero:
Fixes the links to SNApplication and DeviceInterface and makes them relative for better robustness.


mobitar:
Thanks!

PR #188, opened 2020-12-30
feat: display more helpful error message on 403s

baptiste-grob:


lib/services/api/responses.ts
```diff
@@ -11,6 +11,8 @@ export enum StatusCode {
   HttpStatusExpiredAccessToken = 498,
   /** The session's access token and refresh token are expired, user must reauthenticate */
   HttpStatusInvalidSession = 401,
+  /** User's IP is rate-limited. */
+  HttpStatusForbidden = 401,
```
mobitar, line 15: Should be 403 ;)

baptiste-grob: 🤦‍♀️ thanks for the catch

mobitar (approved):

karolsojko (commented):
Can you bump the package version?

PR #191, opened 2021-01-02
chore: add support for node 15

radko93:
As the minimum node version supported natively in arm64 Macs is 15, it would be good if we could add support for it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,8 @@
 {
   "name": "@standardnotes/snjs",
-  "version": "2.0.31",
+  "version": "2.0.32",
   "engines": {
-    "node": ">=14.0.0 <15.0.0"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "@standardnotes/snjs",
-  "version": "2.0.31",
+  "version": "2.0.32",
   "engines": {
-    "node": ">=14.0.0 <15.0.0"
+    "node": ">=14.0.0 <16.0.0"
```
karolsojko, line 5: this still supports 14 as the minimum - should 15 be the minimum as the PR title says?

radko93: True, I changed the title.

karolsojko (approved):

PR #193, opened 2021-01-05
Privileges simplification

baptiste-grob:
I want to get this going quickly because I could see this latest proposed API being another controversial one 😛 so I figured, let's get this discussion going early.

For now please ignore the implementation of the method and instead focus on its signature:

```ts
/**
 * @returns true if the note can be accessed, or a promise resolving
 * to a boolean that says if the note can be accessed or not
 * (taking the form of a challenge)
 */
public authorizeNoteAccess(note: SNNote): true | Promise<boolean>
```

This method would be called every time a user clicks/touches a note in their list, for example.

One might ask, why not just return `Promise<boolean>`?

The problem with promises is: regardless of what they are doing, the specification requires then to resolve on the a future iteration of the event loop. This is the current nature of JavaScript: the result of async work is put in a job queue and will get processed synchronously when there is nothing else to process on this thread. So it's up to the engine's discretion.

However, I think we can all agree that viewing a note is an extremely critical action that should be as fast as possible, instant even (given our constraints or using JS, of course). If a note is not protected, waiting for a whole iteration of the event loop to go by before deciding to show it is wasteful and SNJS should not put that kind of performance constraint on clients, especially because it does not know where it will run.

So while it would make consuming this API a touch dinky:

```ts
const authorized = application.authorizeNoteAccess(note);
if (authorized === true || await authorized) {
  // Show the note
}
```

In my opinion the benefits are worth it: having unprotected note access resolve in the same loop.
Let me know what you think @mobitar @radko93


baptiste-grob:
An alternative API (let's call it API #2) would be to have two methods instead:

```ts
if (application.noteIsAccessible(note) || await application.authorizeNoteAccess(note)) {
  // Show the note
}
```

baptiste-grob:
API 3 could use callbacks and thus keep it in all in one method:

```ts
application.requestNoteAccess(note, (granted) => {
  if (granted) {
    // Show the note
  }
});
```

radko93:
As the action really has to be instant if possible (basically every millisecond matters), I like the idea of returning quickly. Probably v1 is OK for me.

mobitar:
Milliseconds really don’t matter that much, if we’re dealing with like 10 or so. Isn’t the human eye frame rate like 30 or something?

And all clients are async anyway presently. Lots of UI actions depend on an async call and it’s not perceptible.

So I wouldn’t make a convoluted API to shave 1ms off a local interaction.

baptiste-grob:
Milliseconds add up and this is not the only work required to show a note so I would rather we nip this one in the bud and ensure death by a thousand cuts cannot happen.

What do you think of API 2 @mobitar? It's essentially similar to what clients are doing right now with privileges but with updated naming.

mobitar:
> Milliseconds add up

You'd have to add up over a hundred of them before it becomes perceptible. Do some search on the topic. 

This btw would be the literal definition of premature optimization. Making code complex to achieve goals that don't really end up mattering.

In fact even if the cost was 10ms I would never sleep easy with an API like this:

```
if (authorized === true || await authorized) {
  // Show the note
}
```

Option B is the best of the solutions presented for this imaginary problem (😉), but again, if this becomes a pattern across the codebase (over-optimizing async calls), we end up having a really ugly codebase.

In any case `application.noteRequiresAuthorization(note) || await application.authorizeNoteAccess(note)` for function names better ties the two together. Or `doesNoteRequireAuthorization` or something. Just using the same verb.

baptiste-grob:
Well, you clearly feel stronger about this than I do and because it's not a hard change to reverse if it becomes a problem I'll go with the single-promise approach :)

mobitar:
> feel stronger about this than I do

Only because I've been here before and ended up making ugly decisions and over-solving problems that turned out to be only theoretical in nature.

It's always tradeoffs though. Here I would choose a cleaner API over over-worrying about a single event loop cycle. But in our case it's an easy decision because SNJS is nearly 100% async, and so clients will not really be able to escape this reality.

mobitar:
Also note selection is not as critical a UI function performance-wise as keystrokes would be, for example. Saving 1ms off keystrokes could be crucial (one could make the case), but sub-1ms delays after a click are physically indistinguishable.

lib/services/privileges_service.ts
```diff
@@ -71,8 +71,21 @@ export class SNPrivilegesService extends PureService {
...
 
+    return this.validateOrRenewSession(ChallengeReason.AccessProtectedNote);
+  }
+
+  async authorizeFileImport(): Promise<boolean> {
+    return this.validateOrRenewSession(ChallengeReason.ImportFile);
+  }
+
+  async authorizeAutolockIntervalChange(): Promise<boolean> {
+    return this.validateOrRenewSession(ChallengeReason.ImportFile);
```
mobitar, line 82: Copy/paste error here..

mobitar:
Should we be targeting develop branch instead?

mobitar:
Oh this is SNJS, nm. Thought I was in web 😄

lib/application.ts
```diff
@@ -1388,22 +1405,39 @@ export class SNApplication {
     }
   }
 
-  public async removePasscode(): Promise<void> {
+  /**
+   * @returns whether the passcode was successfuly removed or not
+   */
+  public async removePasscode(): Promise<boolean> {
+    const passcode = await this.challengeService.promptForPasscode(
```
mobitar, line 1412: I feel like `challengeService.promptForPasscode` should be renamed to `challengeService.promptForCorrectPasscode`, if that's the way it works. Not necessary for this PR but something to keep in mind as reading this code is confusing because we don't even validate the resulting `passcode`.

lib/application.ts
```diff
@@ -1388,22 +1405,39 @@ export class SNApplication {
...
   public async changePasscode(
-    passcode: string,
+    newPasscode: string,
     origination = KeyParamsOrigination.PasscodeChange
-  ): Promise<void> {
+  ): Promise<boolean> {
+    const passcode = await this.challengeService.promptForPasscode(
+      ChallengeReason.ChangePasscode
+    );
+    if (!passcode) return false;
```
mobitar, line 1439: On line 1415 you use `if (isNullOrUndefined(passcode)) return false;` but here it's a direct check. Not sure if there's a practical difference.

lib/models/app/privileges.ts
```diff
@@ -22,6 +22,9 @@ type PrivilegeMap = Partial<Record<ProtectedAction, PrivilegeCredential[]>>
 /**
  * Privileges are a singleton object that store the preferences a user
  * may have configured for protecting certain actions.
+ *
+ * @deprecated Privileges are now inferred based on the user's active
```
mobitar, line 26: Could we just delete this file? Don't think we need it..

lib/services/privileges_service.ts
```diff
@@ -1,192 +1,173 @@
...
- * For example, the privileges service exposes functions that allow the action of deleting
- * a note or viewing a note to require extra authentication.
- * Privileges are a superficial-level locking feature; they do not deal at all with underlying
- * data state. For example, viewing a protected note may require extra authentication,
- * but the underlying note data may already be decrypted in application memory, whether or not
- * the user has yet authenticated this action.
+ * Privileges enforce certain actions to require extra authentication,
+ * like viewing a protected note.
  */
 export class SNPrivilegesService extends PureService {
```
mobitar, line 53: I would be ok renaming this class to SNProtectionService, if you also dislike the word privileges as much as I do. But perhaps can be done in a followup PR so that the rename doesn't make git think it's a new file.

test/keys.test.js
```diff
@@ -22,7 +22,7 @@ describe('keys', function () {
   });
 
   afterEach(function () {
-    this.application.deinit();
+    this.application?.deinit();
```
mobitar, line 25: TS in JS file?

test/keys.test.js
```diff
@@ -408,6 +408,17 @@ describe('keys', function () {
...
+    this.application.setLaunchCallback({
+      receiveChallenge: (challenge) => {
+        this.application.submitValuesForChallenge(
+          challenge,
+          challenge.prompts.map(prompt => new ChallengeValue(prompt, passcode))
+        )
+      }
+    });
+
     await this.application.setPasscode('foo');
```
mobitar, line 422: Should use new `passcode` const

lib/application.ts
baptiste-grob: There is none but the intent is clearer. I'll amend this

test/keys.test.js
baptiste-grob: It is JS (and thus valid TS) https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

lib/application.ts
baptiste-grob: There's no difference but it makes the check clearer so I will amend this change

lib/services/privileges_service.ts
baptiste-grob: I agree, the word privileges have a lot of unnecessary connotations here… I'll make this change in a separate PR. How about `SNAuthorizationService` instead, to align it with the methods in this file? (And also because it only protects something if the client is nice enough to call the methods 😄)

lib/models/content_types.ts
```diff
@@ -13,6 +13,10 @@ export enum ContentType {
   Editor = 'SN|Editor',
   ActionsExtension = 'Extension',
   UserPrefs = 'SN|UserPreferences',
+  /**
+   * @deprecated Privileges are now inferred based on the user's active
+   * protections, like passcode or biometrics
+   */
   Privileges = 'SN|Privileges',
```
mobitar, line 20: Couldn't we just delete this content type?

lib/services/privileges_service.ts
mobitar: That could work. `auth_service` is what used to be called the service to auth with the server. So it could go either way I guess. Could the remaining logic in the privileges_service just be grouped into the challenge_service? (Since ultimately a challenge is what's occurring or not occurring.)

radko93 (approved):

mobitar (approved):

lib/models/content_types.ts
mobitar: Probably should be deleted actually so that clients fail to compile rather than a warning. If they're using this they shouldn't be.

baptiste-grob: I think so but I'm not completely sure. For now I'd like to keep it as documentation and make a decision later.

mobitar: GitHub is our source of documentation. I honestly don't think it's worth keeping around or postponing this.

lib/services/privileges_service.ts
baptiste-grob: I'd like to keep the responsibility of handling authorization session length in a separate class, as that part is completely unrelated to the challenge_service

PR #197, opened 2021-01-11
fix: return errored payload instead of throwing

mobitar:


mobitar:
For issue https://github.com/standardnotes/forum/issues/1197

baptiste-grob (approved):

PR #198, opened 2021-01-11
feat: destroy revoked session data

baptiste-grob:


lib/application.ts
```diff
@@ -1573,6 +1574,12 @@ export class SNApplication {
 
   private createHttpManager() {
     this.httpService = new SNHttpService();
+    this.serviceObservers.push(this.httpService.addEventObserver((tag) => {
+      if (tag === ErrorTag.RevokedSession) {
+        /** Sign out without warning. */
```
mobitar, line 1579: Hmm, thinking about "without warning"..

baptiste-grob: I think in the end if you're revoking a session you could be defending yourself against an attacker, and giving them a warning before destroying the data would give them a window to back it up, thus defeating the purpose of the destruction mechanism.

Then maybe revoking a session needs to be an authenticated action (as in, prompts for password/passcode/biometrics).

baptiste-grob:
Complete feature-wise, but still needs tests

lib/services/api/messages.ts
```diff
@@ -131,7 +131,9 @@ export const SessionStrings = {
     'The associated app will be signed out and all data removed ' +
     'from the device when it is next launched. You can sign back in on that ' +
     'device at any time.',
-}
+  CurrentSessionRevoked:
+    'Your session has been revoked and all local data has been destroyed.',
```
mobitar, line 135: Let's go with "...has been removed from this device." so it sounds less scary.

test/session.test.js
```diff
@@ -597,11 +597,10 @@ describe('server session', function () {
...
     await this.application.revokeSession(app2session.uuid);
-
-    const note = await Factory.createSyncedNote(app2);
-    await this.application.sync();
-    expect(this.application.findItem(note.uuid)).to.not.be.ok
-    app2.deinit();
+    void app2.sync();
+    /** Wait for app2 to deinit */
+    await Factory.sleep(3);
+    expect(app2.dealloced).to.be.true;
```
mobitar, line 603: It seems we’re testing that the app has been successfully destroyed, but not whether local data has been deleted. Because an app can be dealloced and still hang on to data on storage. So this test isn’t fully comprehensive I think.

docker-compose.yml
```diff
@@ -82,7 +83,7 @@ services:
       LOG_LEVEL: debug
       NODE_ENV: development
       JWT_SECRET: s3cr3tK3y
-      ACCESS_TOKEN_AGE: 2
+      ACCESS_TOKEN_AGE: 4
```
karolsojko, line 86: 👍

mobitar (approved):

PR #204, opened 2021-01-19
feat: typedoc

TheodoreChu:


README.md
```diff
@@ -132,5 +132,21 @@ yarn test
...
+```
+yarn docs
+```
+
+This will generate documentation and place it in the `docs/` directory. Create a localhost server using `http-server` and open the [docs/index.html](docs/index.html) file in a browser to view the docs.
+
+To deploy the latest version of the docs to the `gh-pages` branch to publish using [GitHub Pages](https://pages.github.com/), replace `2.0.41` with the latest version and run:
+
+```
+yarn deploy-docs -m "build: v2.0.41"
```
karolsojko, line 148: Can this be a part of the Github Actions workflow when we are releasing a new version? See: https://github.com/standardnotes/snjs/blob/master/.github/workflows/publish.yml

karolsojko (changes requested):

README.md
TheodoreChu: Yes, it could be part of the GitHub Actions workflow when we release a new version or a separate workflow after a new push. Which do you think is better? Do you want me to figure out how to add the new workflow or do you want to do it?

karolsojko: should be in the `publish` workflow right after the package is published we should add a job for publishing docs

TheodoreChu: That sounds good

TheodoreChu: Do you want me to update the publish workflow or do you want to do it?

karolsojko: please update it if you can - if it's too troublesome I'll attend it at a later stage

PR #210, opened 2021-01-22
feat: add config package with common configuration files

karolsojko:


packages/config/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "version": "1.0.0",
+  "engines": {
+    "node": ">=14.0.0 <16.0.0"
+  },
+  "description": "Common configuration files used in SN projects",
+  "publishConfig": {
+    "access": "public"
+  },
+  "author": "Karol Sójko <karolsojko@gmail.com>",
+  "license": "MIT",
```
mobitar, line 12: Let's go with AGPLv3 for everything

karolsojko: Will do :)

karolsojko: @mobitar updated

mobitar (approved):

PR #214, opened 2021-02-08
feat: add Auth SDK package

karolsojko:


packages/auth/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "description": "Auth SDK for Standard Notes projects",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "author": "Karol Sójko <karolsojko@gmail.com>",
```
baptiste-grob, line 16: Personal email?

baptiste-grob (approved):

packages/auth/package.json
karolsojko: hmm I assumed it has to be the one which I use as a github account

karolsojko: will update in that case

PR #215, opened 2021-02-08
feat: add .prettierrc to config package

TheodoreChu:
I think this matches `.eslintrc`: https://github.com/standardnotes/snjs/blob/da023dd2d3568b972c5eed259067375bb0505669/packages/config/src/.eslintrc#L20


baptiste-grob (commented):
Great, thanks! Can you also update the package.json version to 1.1.0 before merging?

baptiste-grob (approved):

PR #216, opened 2021-02-11
feat(wip): auth & permissions

baptiste-grob:


baptiste-grob:
Since SNJS is supposed to behave identically to the previous version, no test has been added.
Tests will be written when a subsequent PR that actually makes use of permissions will be created, which will happen after we've updated the testing infra to use the new API gateway.

packages/snjs/lib/services/api/api_service.ts
```diff
@@ -113,104 +151,117 @@ export class SNApiService extends PureService {
...
-    if (!host) {
-      throw `Attempting to build path ${path} with no host.`;
-    }
-    if (!path) {
-      throw 'Attempting to build path with null path.';
-    }
-    return joinPaths(host, path);
-  }
-
-  private get apiVersion() {
```
mobitar, line 127: Needed for this test: https://github.com/standardnotes/snjs/blob/master/test/keys.test.js#L820

baptiste-grob: Thanks! I fixed it by making `V0_API_VERSION` a static class variable.

packages/snjs/lib/application.ts
```diff
@@ -152,9 +154,11 @@ export class SNApplication {
   private itemManager!: ItemManager;
   private keyRecoveryService!: SNKeyRecoveryService;
   private preferencesService!: SNPreferencesService;
+  private permissionsService!: SNPermissionsService;
+
 
   private eventHandlers: ApplicationObserver[] = [];
-  private services: PureService<any>[] = [];
+  private services: PureService<any, any>[] = [];
```
standarius, line 161: why not `unknown` instead of `any`?

packages/snjs/lib/services/api/http_service.ts
```diff
@@ -125,7 +125,9 @@ export class SNHttpService extends PureService {
     try {
       if (httpStatus !== StatusCode.HttpStatusNoContent) {
         const body = JSON.parse(request.responseText);
-        response.object = body;
+        if (!body.data) {
+          response.data = body;
+        }
         Object.assign(response, body);
```
standarius, line 131: So `response` might contain `body` in its `data` field (via the added `if`) and also all of `body`'s properties at the same time. Is that ok?

packages/snjs/lib/services/pure_service.ts
```diff
@@ -2,32 +2,36 @@ import { removeFromArray } from '@Lib/utils';
...
 
-type EventObserver<T> = (eventName: T, data: any) => Promise<void> | void;
+type EventObserver<E, D> = (eventName: E, data?: D) => Promise<void> | void;
 
-export abstract class PureService<E = string> {
-  private eventObservers: EventObserver<E>[] = [];
+/**
+ * E = the name of the event
+ * D = the data the event includes
+ */
```
standarius, line 10: Perhaps use descriptive generic type names and remove the comment instead?

packages/snjs/lib/application.ts
baptiste-grob: With unknown I wouldn't be able to push services into this array without an explicit cast, which I didn't think was much of a better solution… If you've got an idea on how to make this better please let me know

packages/snjs/lib/services/api/http_service.ts
baptiste-grob: `body` is the whole request body, not a specific key.
Under the future api, response bodies will look closer to  be contained in a root-level `data` key (there will also be a `meta` key, to move closer to a [JSON:API](https://jsonapi.org/)-complient API shape). This code is here to handle legacy responses that may not have a data key, allowing consumers of the httpService to treat everything the same way regardless of the response shape.
Your remark makes me realize that it should be explained somehow!

packages/snjs/lib/services/api/api_service.ts
```diff
@@ -2,50 +2,78 @@ import { UuidString } from './../../types';
...
 
 type InvalidSessionObserver = (revoked: boolean) => void;
 
 export class SNApiService extends PureService {
-  private host?: string;
+  /** Legacy api version field to be specified in params when calling v0 APIs. */
+  static readonly V0_API_VERSION = '20200115';
+
   private session?: Session;
+  private host!: string;
```
standarius, line 76: is the bang operator necessary here?

packages/snjs/lib/services/api/http_service.ts
baptiste-grob: I've pushed an update with a comment, let me know if that makes things clearer :)

standarius: I see. So we duplicate the data for compatibility.

packages/snjs/lib/services/api/api_service.ts
baptiste-grob: Hm, I should probably just make `this.host` a required parameter in the constructor yeah…

packages/snjs/lib/application.ts
standarius: I see. No quick solution I can figure out.

packages/auth/tsconfig.json
```diff
@@ -3,5 +3,8 @@
   "compilerOptions": {
     "rootDir": "src",
     "outDir": "dist"
-  }
+  },
+  "exclude": [
+    "dist"
```
karolsojko, line 8: not sure but hmm maybe this should go for all packages in the `@standardnotes/config`?

packages/snjs/lib/services/api/api_service.ts
```diff
@@ -2,49 +2,76 @@ import { UuidString } from './../../types';
...
+  sync: string;
+  signOut: string;
+  refreshSession: string;
+  sessions: string;
+  session: string;
+  itemRevisions: (itemId: string) => string;
+  itemRevision: (itemId: string, revisionId: string) => string;
+};
+
+const Paths: {
```
karolsojko, line 50: cool idea 👍

karolsojko (approved):

packages/auth/tsconfig.json
baptiste-grob: Good point!

PR #217, opened 2021-02-11
fix: protections

baptiste-grob:
- [x] Require local (no password) authentication when exporting a decrypted backup
- [x] Expose session expiry date
- [x] Expose clearing session
- [x] Require authentication when disabling biometrics
- [x] Have the same prompt order when launching the app as other authenticated actions


karolsojko (approved):

packages/snjs/lib/device_interface.ts
```diff
@@ -34,7 +34,7 @@ export abstract class DeviceInterface {
     this.interval = null;
   }
 
-  abstract async getRawStorageValue(key: string): Promise<string | undefined>;
+  abstract getRawStorageValue(key: string): Promise<string | undefined>;
```
mobitar, line 37: Why is async removed? React Native uses AsyncStorage, which is fully async.

mobitar: Oh nevermind, back to the return type vs annotation argument 😓

baptiste-grob: In this case it's because TypeScript marks it as an error 😅 I don't know

packages/snjs/lib/services/history/history_manager.ts
```diff
@@ -74,11 +74,9 @@ export class SNHistoryManager extends PureService {
     this.persistable = await this.storageService!.getValue(
       StorageKey.SessionHistoryPersistable
     );
-    this.sessionHistory = await this.storageService!.getValue(
-      StorageKey.SessionHistoryRevisions
-    ).then((historyValue) => {
-      return SessionHistoryMap.FromJson(historyValue);
-    });
+    this.sessionHistory = SessionHistoryMap.FromJson(
+      this.storageService!.getValue(StorageKey.SessionHistoryRevisions)
```
mobitar, line 78: How does this work? getValue is async no?

packages/snjs/lib/device_interface.ts
baptiste-grob: https://stackoverflow.com/questions/65355164/error-ts1243-async-modifier-cannot-be-used-with-abstract-modifier

packages/snjs/lib/services/history/history_manager.ts
mobitar: Ah nm, it's not async.

baptiste-grob: Maybe getValue used to read something from disk? In any case it's actually completely sync (reads from memory) and I removed the modifier to address a lint warning. This refactor is in response to the compile error that followed

PR #223, opened 2021-02-16
feat: add isSameDay util

antsgar:
Add isSameDay util that returns a boolean indicating whether two dates correspond to the same day. This util is moved here because it's used both for web and mobile.

### DIFF
--- a/test/utils.test.js
+++ b/test/utils.test.js
@@ -159,7 +159,7 @@ describe('utils', () => {
   });
 
   describe('isSameDay', () => {
-    it('Returns true if two dates are on the same day', () => {
+    it('returns true if two dates are on the same day', () => {
       const dateA = new Date(2021, 1, 16, 16, 30, 0);
       const dateB = new Date(2021, 1, 16, 17, 30, 0);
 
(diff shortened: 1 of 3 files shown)


baptiste-grob:
Cool, thanks :) Could you add one of two tests for it? They would fit well into the existing `utils.test.js` file and is probably the smoothest way you can get into SNJS testing 😄 please don't hesitate to DM me if you need more specific info/guidance

antsgar:
> Cool, thanks :) Could you add one of two tests for it? They would fit well into the existing `utils.test.js` file and is probably the smoothest way you can get into SNJS testing 😄 please don't hesitate to DM me if you need more specific info/guidance

Ah yes, sorry, I missed the test file! Will do 🙂

test/utils.test.js
```diff
@@ -157,4 +157,22 @@ describe('utils', () => {
     const result = truncateHexString(hex256, desiredBits);
     expect(result.length).to.equal(expectedLength);
   });
+
+  describe('isSameDay', () => {
+    it('Returns true if two dates are on the same day', () => {
```
mobitar, line 162: Not a huge deal but at least for our test suite we don't uppercase the first letter of the sentence

baptiste-grob (approved):

PR #224, opened 2021-02-17
feat: SNApplication.hasProtections()

baptiste-grob:


packages/snjs/lib/application.ts
```diff
@@ -1400,6 +1400,10 @@ export class SNApplication {
     return this.launched;
   }
 
+  public hasProtections(): boolean {
```
mobitar, line 1403: Our Application interface should be the documentation hub for all clients. So I think it's worth putting a function description here that clients can read as they hover over the function. What does hasProtections mean? What is it used for?

Also I think more descriptive naming could help with documentation. `hasProtectionSources` is more self-evident.

Then you'd also have another function, `isProtectionsEnabled` that equals `hasProtectionSources() && protectionExpirey >/< some other date`

mobitar: Also don't tell Baptiste but he would surely reject this PR for a new function without tests ;)

baptiste-grob: busted

PR #225, opened 2021-02-17
protections public methods

baptiste-grob:
With this PR we'll end up with:

- `SNApplication.hasProtections()`: to determine if a user can use the feature at all (which they can't if they don't have an account, passcode, or biometrics authentication)
- `SNApplication.areProtectionsEnabled()`: to determine if protections are enabled/disabled

### DIFF
--- a/packages/snjs/lib/application.ts
+++ b/packages/snjs/lib/application.ts
@@ -889,8 +889,8 @@ export class SNApplication {
   }
 
   /**
-   * If a user does not have any protection source it may not be relevant to
-   * show them any information related to them.
+   * Use this method if you do not want to display protections-related
+   * information to a user who cannot currently exploit the feature.
    */
   public hasProtectionSources(): boolean {
     return this.protectionService.hasProtectionSources();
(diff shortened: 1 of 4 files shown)


mobitar:
Left some comments here: https://github.com/standardnotes/snjs/pull/224/files#r577674680

baptiste-grob:
Should be good to review now @mobitar @antsgar :)

packages/snjs/lib/application.ts
```diff
@@ -888,6 +888,22 @@ export class SNApplication {
     return this.protocolService.hasAccount();
   }
 
+  /**
+   * Use this method if you do not want to display protections-related
+   * information to a user who cannot currently exploit the feature.
+   */
+  public hasProtectionSources(): boolean {
```
mobitar, line 895: What do you think of a comment like this:

```
Returns true if the user has a source of protection available, such as a passcode, password, or biometrics.
```

I find the current comment somewhat awkwardly phrased. "Use..if you do not..who cannot..exploit"..

antsgar (approved):

packages/snjs/lib/application.ts
baptiste-grob: Sounds good! I was struggling with this one

PR #227, opened 2021-02-22
fix: eagerly load preferences

baptiste-grob:
Currently preferences are only loaded after a full sync event has completed, which is much later than when a user can interact with their notes

### DIFF
--- a/packages/snjs/lib/services/preferences_service.ts
+++ b/packages/snjs/lib/services/preferences_service.ts
@@ -1,4 +1,4 @@
-import { ContentType, SNPredicate, SNUserPrefs } from '@Lib/models';
+import { ContentType, SNUserPrefs } from '@Lib/models';
 import {
   PrefKey,
   PrefValue,
@@ -51,7 +51,9 @@ export class SNPreferencesService extends PureService<'preferencesChanged'> {
   public async handleApplicationStage(stage: ApplicationStage): Promise<void> {
     await super.handleApplicationStage(stage);
     if (stage === ApplicationStage.LoadedDatabase_12) {
-      void this.reload();
+      this.preferences = this.singletonManager.findSingleton(
+        SNUserPrefs.singletonPredicate
+      );
     }
   }
 
@@ -88,12 +90,10 @@ export class SNPreferencesService extends PureService<'preferencesChanged'> {
     }
     this.reloading = true;
     try {
-      const contentType = ContentType.UserPrefs;
-      const predicate = new SNPredicate('content_type', '=', contentType);
       const previousRef = this.preferences;
       this.preferences = await this.singletonManager.findOrCreateSingleton<SNUserPrefs>(
-        predicate,
-        contentType,
+        SNUserPrefs.singletonPredicate,
+        ContentType.UserPrefs,
         FillItemContent({})
       );
       if (
--- a/packages/snjs/lib/services/preferences_service.ts
+++ b/packages/snjs/lib/services/preferences_service.ts
@@ -12,7 +12,10 @@ import { SNSyncService } from './sync/sync_service';
 import { SyncEvent } from './sync/events';
 import { ApplicationStage } from '@Lib/stages';
 
-export class SNPreferencesService extends PureService<'preferencesChanged'> {
+const preferencesChangedEvent = 'preferencesChanged';
+type PreferencesChangedEvent = typeof preferencesChangedEvent;
+
+export class SNPreferencesService extends PureService<PreferencesChangedEvent> {
   private shouldReload = true;
   private reloading = false;
   private preferences?: SNUserPrefs;
@@ -51,9 +54,13 @@ export class SNPreferencesService extends PureService<'preferencesChanged'> {
   public async handleApplicationStage(stage: ApplicationStage): Promise<void> {
     await super.handleApplicationStage(stage);
     if (stage === ApplicationStage.LoadedDatabase_12) {
+      /** Try to read preferences singleton from storage */
       this.preferences = this.singletonManager.findSingleton(
         SNUserPrefs.singletonPredicate
       );
+      if (this.preferences) {
+        void this.notifyEvent(preferencesChangedEvent);
+      }
     }
   }
 
@@ -80,7 +87,7 @@ export class SNPreferencesService extends PureService<'preferencesChanged'> {
         m.setPref(key, value);
       }
     )) as SNUserPrefs;
-    void this.notifyEvent('preferencesChanged');
+    void this.notifyEvent(preferencesChangedEvent);
     void this.syncService.sync();
   }
 
--- a/packages/snjs/lib/services/preferences_service.ts
+++ b/packages/snjs/lib/services/preferences_service.ts
@@ -36,7 +36,7 @@ export class SNPreferencesService extends PureService<PreferencesChangedEvent> {
       }
     );
 
-    this.removeSyncObserver = syncService.addEventObserver(async (event) => {
+    this.removeSyncObserver = syncService.addEventObserver((event) => {
       if (event === SyncEvent.FullSyncCompleted) {
         void this.reload();
       }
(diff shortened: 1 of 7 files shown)


mobitar:
Wouldn't this overwrite the server value? The reason we previously waited for full sync to complete is so that we're sure the server doesn't have a value before we create our own. If you reload before server has loaded, the singleton manager will create a new preferences item. (I think)

One solution could be to read-only before server value has loaded, then read/create after we're synced up with the server.

baptiste-grob:
Good point, in theory this is an unlikely scenario, since when signing in for the first time the server singleton takes priority, so unless the app is closed before the first sync completes (and the singleton gets stored) this situation would never occur as the singleton would get read from storage each time.

However it looks like there is no existing test to assert what should happen when you're creating a singleton before a full sync has completed outside of a signin flow, so I'll work on implementing that.

baptiste-grob:
Wrote the aforementioned test which passes on my local machine 👍 please review

packages/snjs/lib/services/preferences_service.ts
```diff
@@ -46,6 +47,12 @@ export class SNPreferencesService extends PureService<'preferencesChanged'> {
     (this.itemManager as unknown) = undefined;
   }
 
+  public handleApplicationStage(stage: ApplicationStage): void {
+    if (stage === ApplicationStage.LoadedDatabase_12) {
```
mobitar, line 51: We should probably call super here as well.

packages/snjs/lib/services/preferences_service.ts
```diff
@@ -46,6 +47,12 @@ export class SNPreferencesService extends PureService<'preferencesChanged'> {
     (this.itemManager as unknown) = undefined;
```
mobitar, line 47: Should probably be calling super here as well

test/singletons.test.js
```diff
@@ -240,6 +240,37 @@ describe('singletons', function() {
...
+    this.application = await Factory.createAndInitializeApplication(identifier);
+
+    /**
+     * After restarting and syncing, the instance retrieved from the server
+     * should be the one kept
+     */
+    const latestPrefs = await insertPrefsPayload(this.application);
+    expect(latestPrefs.uuid).to.equal(ogPrefs.uuid);
+    expect(this.application.findItem(localPrefs.uuid)).to.not.exist;
+    const allPrivs = this.application.itemManager.nonErroredItemsForContentType(ogPrefs.content_type);
```
mobitar, line 270: Should be allPrefs

test/singletons.test.js
```diff
@@ -240,6 +240,37 @@ describe('singletons', function() {
...
+
+    await Factory.sleep(0.05); /** Give the application time to save payloads */
+    await this.application.deinit();
+    this.application = await Factory.createAndInitializeApplication(identifier);
+
+    /**
+     * After restarting and syncing, the instance retrieved from the server
+     * should be the one kept
+     */
+    const latestPrefs = await insertPrefsPayload(this.application);
```
mobitar, line 267: Can we rename `insertPrefsPayload` to `findOrCreatePrefsSingleton`. Kind of confusing reading this code and trying to understand it with that function name.

test/singletons.test.js
```diff
@@ -240,6 +240,37 @@ describe('singletons', function() {
...
+    await this.registerUser();
+
+    /** Create prefs and associate them with account */
+    const ogPrefs = await insertPrefsPayload(this.application);
+    await this.application.sync(syncOptions);
+
+    /** Remove prefs locally and create a newer one */
+    this.application.lockSyncing();
+    await this.application.itemManager.removeItemLocally(ogPrefs);
+    const localPrefs = await insertPrefsPayload(this.application);
```
mobitar, line 254: If syncing is locked this value would never be persisted to the DB. So when you recreate the application `expect(this.application.findItem(localPrefs.uuid)).to.not.exist;` would always pass.

baptiste-grob: Ah shoot, so my test may be bonked then. Is there an easy way to persist an item to storage without syncing that you know of?

mobitar: storageService.savePayload probably

mobitar: Just to be sure you should also add an assertion after you deinit that the item can be found in storage.

baptiste-grob: Awesome, thanks!

baptiste-grob:
Now that I have a correct test, it turns out the answer to your question:

> Wouldn't this overwrite the server value?

Is yes. So instead of defining additional special behavior for singletons, which would increase the general complexity of the system for not much benefit, I've flipped the test around and changed the preferences service to try to find, but not create, a preferences singleton on startup. Let me know what you think!

test/singletons.test.js
```diff
@@ -222,22 +223,54 @@ describe('singletons', function() {
...
+  it('keeps local singleton when it is created before sync', async function () {
+    const identifier  = this.application.identifier;
+    await this.registerUser();
+
+    /** Create prefs and associate them with account */
+    const ogPrefs = await findOrCreatePrefsSingleton(this.application);
+    await this.application.sync(syncOptions);
+
+    /** Remove prefs locally and create a newer one */
+    await this.application.storageService.deletePayloads([ogPrefs]);
```
mobitar, line 253: Hmm, ogPrefs is an Item, but deletePayloads takes in a Payload, not an Item. You can probably use ogPrefs.payload.

mobitar:
Nice, always good to have simpler solutions 👌 Btw you should always write failing tests first. I think that would have caught this issue. If you write a test but never see it fail you really can't prove to yourself it's written correctly.

test/singletons.test.js
```diff
@@ -222,22 +223,54 @@ describe('singletons', function() {
...
-    const latestPrefs = await insertPrefsPayload(this.application);
+    const latestPrefs = await findOrCreatePrefsSingleton(this.application);
     expect(latestPrefs.uuid).to.equal(ogPrefs.uuid);
-    const allPrivs = this.application.itemManager.nonErroredItemsForContentType(ogPrefs.content_type);
-    expect(allPrivs.length).to.equal(1);
+    const allPrefs = this.application.itemManager.nonErroredItemsForContentType(ogPrefs.content_type);
+    expect(allPrefs.length).to.equal(1);
+  });
+
+  it('keeps local singleton when it is created before sync', async function () {
```
mobitar, line 244: I'm really not sure what this test proves anymore. I think here you seem to be just testing storage behavior rather than anything to do with singletons: if you delete something from storage and replace it with another entity, then surely when the app initializes, it will only have the new one, as it's like old one never even existed if you surreptitiously delete it from storage.

Can you describe what you're trying to test? How is `keeps local singleton when it is created before sync` describing the patch you made?

baptiste-grob: It's not a test that is related to the PR anymore, but rather one that proves that a server singleton does not have priority outside of downloadFirstSync. That way if this behavior ever changes the test will fail, prompting us to question whether this was intended or not.

You're right that I do need to write another test for this specific PR, though 😅

mobitar: If you want to prove that an incoming server item will not override the local copy, then you need to clear the sync token and sync again from scratch. Otherwise I'm not sure this test proves anything..

test/preferences.test.js
```diff
@@ -92,4 +92,37 @@ describe('preferences', function () {
...
+
+    this.application = Factory.createApplication(identifier);
+
+    let preferences;
+    const removeEventObserver = this.application.addEventObserver(() => {
+      preferences = this.application.preferencesService.preferences;
+      removeEventObserver();
+    }, ApplicationEvent.WillSync);
+
+    return new Promise((resolve, reject) => {
```
mobitar, line 115: Seems like this all could be simplified to this?

```
    return new Promise((resolve, reject) => {
      let preferences;
      const removeEventObserver = this.application.addEventObserver(() => {
        preferences = this.application.preferencesService.preferences;
        removeEventObserver();
        if (preferences) {
          expect(this.application.getPreference(prefKey)).to.equal(prefValue);
          resolve();
        } else {
          reject(Error('preferences have not been loaded.'));
        }
      }, ApplicationEvent.WillSync);

      Factory.initializeApplication(this.application);
    })
```

mobitar (approved):

PR #233, opened 2021-03-04
fix: broken references to notes in tags when importing backup

standarius:
Related issue: https://app.asana.com/0/1160981203938995/1199383179826917/f

This fixes the problem of disappearing tags: https://github.com/standardnotes/forum/issues/1179

On my machine the backup file from this issue (https://github.com/standardnotes/forum/files/5745950/backup.zip) imports properly with the correct number of tags and correct references to all notes.

It's a quick fix (first hack that works). I cannot guarantee it doesn't break something else as there is seems to be a lot of dependencies and state in this code and I had no time to understand them.

***

I just noticed that after rebasing to master, there are some errors, probably caused by latest commits. Will fix that later.

In the meantime leaving as is.

### DIFF
--- a/packages/snjs/lib/protocol/payloads/deltas/remote_conflicts.ts
+++ b/packages/snjs/lib/protocol/payloads/deltas/remote_conflicts.ts
@@ -6,6 +6,9 @@ import { ImmutablePayloadCollection } from "@Protocol/collection/payload_collect
 import { PayloadsByAlternatingUuid } from '@Payloads/functions';
 import { extendArray } from '@Lib/utils';
 import { PurePayload } from '../pure_payload';
+import { PayloadContent } from '@Lib/protocol';
+import { CopyPayload } from '../generator';
+import { ContentType } from '@Lib/models/content_types';
 
 export class DeltaRemoteConflicts extends PayloadsDelta {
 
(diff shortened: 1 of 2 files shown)


packages/snjs/lib/protocol/payloads/deltas/remote_conflicts.ts
```diff
@@ -69,8 +142,7 @@ export class DeltaRemoteConflicts extends PayloadsDelta {
         PayloadSource.DecryptedTransient
       );
```
standarius, line 143: (commenting on line 139)

`mostRecent` most likely does not work as intended and is the cause of duplicated tags.

mobitar:
So I haven't reviewed the whole thing yet, but there's something important to note:

Payloads are largely abstract. There is no NotePayload or TagPayload. All payloads behave the same. So, logic inside of deltas/remote_conflict for example, should not create custom behavior for notes or tags. There should not be any conditionals that treat tags or notes differently than any other type.

When you want to begin treating tags and notes with special behavior, that's where Items come into play. Items are made out of Payloads, and there are SNNote items and SNTag items.

While this may fix the issue, our goal is to fix the design always, which should subsequently fix the issue.

Secondly, this PR doesn't contain any new or modified tests. In this case, you'd want to first create a failing test case that captures this issue, and ensure that your proposed solution causes this test to pass.

packages/snjs/lib/protocol/payloads/deltas/remote_conflicts.ts
```diff
@@ -56,7 +59,77 @@ export class DeltaRemoteConflicts extends PayloadsDelta {
    */
   private async collectionsByHandlingUuidConflicts() {
     const results: Array<PurePayload> = [];
-    for (const payload of this.applyCollection.all()) {
+
+    const notes = []
+    const tags = []
+    const rest = []
+    for (const item of this.applyCollection.all()
+    ) {
```
standarius, line 67: accidental newline to be removed

mobitar:
Internally discussed to focus on a more fundamental fix.

PR #235, opened 2021-03-06
feat: refactor part of utils

SNCubes:


packages/snjs/lib/utils.ts
```diff
@@ -36,7 +36,7 @@ export function isWebCryptoAvailable() {
 /**
  * Whether we are in React Native app
  */
-export function isReactNativeEnvironment() {
+export function isReactNativeEnvironment() : boolean {
```
mobitar, line 39: Incorrect spacing, should be using auto-linter. It's option + shift + f on Mac also to perform auto-format.

packages/snjs/lib/utils.ts
```diff
@@ -75,7 +74,7 @@ export function concatArrays(...args: any[]) {
 /**
  * @returns Whether the value is a function or object
  */
-export function isObject(value: any) {
+export function isObject<T>(value: T|null): boolean {
```
mobitar, line 77: Incorrect spacing in T|null

SNCubes: Thanks! I had to turn it off because the file was getting distracting. I'll turn it back on and run it.

mobitar:
Hmm not sure why it's counting my commits as changes. This would make it impossible to review your changes in isolation. If we can't figure that out I would recommend creating a new PR by cherry-picking your commits.

baptiste-grob:
Yeah you'll want to merge `master` into your branch :)

PR #237, opened 2021-03-09
fix: require proper component manager permissions

johnny243:
Fixes an issue where granting the `stream-items` permission will let you use the following actions:

- `createItems`
- `saveItems`
- `deleteItems`
- `duplicateItem`


johnny243:
After discussing this with @mobitar, I found out this is the expected behavior.

PR #238, opened 2021-03-10
feat: add new infra setup to SNJS test suite

karolsojko:
Added the infra setup that resembless 1-to-1 our current setup on prod. 

While snjs is contacting the syncing server via port `3123` (some requests are redirected to syncing server js), on port `3124` we have the API Gateway and new api exposed. 

The new API should be utilized in future tests when we switch endpoints one by one to utilize it. 

Additionally added fixed versions of docker images in docker-compose so pulling the latest doesn't break the current setup.


mobitar (approved):

docker-compose.yml
```diff
@@ -71,38 +59,41 @@ services:
       - snjs
 
   syncing-server-js:
-    depends_on:
-      - syncing-server-proxy
-      - syncing-server-ruby
-    image: standardnotes/syncing-server-js
-    entrypoint: ["/tools/wait-for.sh", "syncing-server-ruby", "3000", "./docker/entrypoint.sh", "start-web"]
+    image: standardnotes/syncing-server-js:1.15.11
+    entrypoint: ["./wait-for.sh", "syncing-server-ruby", "3000", "./docker/entrypoint.sh", "start-web"]
```
standarius, line 63:
```
"./wait-for.sh", "syncing-server-ruby"
```
takes care of the `depends_on` here, right?

karolsojko: `depends_on` just waits for the container to start .. doesn't wait for the actual application in the container to be started

docker/syncing-server.env.sample
```diff
@@ -0,0 +1,35 @@
...
+DB_PORT=3306
+DB_DATABASE="testdb"
+DB_USERNAME="testuser"
+DB_PASSWORD="testpassword"
+
+REDIS_URL="redis://cache:6379"
+
+EXTENSIONS_SERVER="http://extensions:3000"
+
+SALT_PSEUDO_NONCE="change-me-!s"
```
standarius, line 20: why the trailing `s` in the value?

test/proxy.conf.template
```diff
@@ -16,18 +16,18 @@ server {
...
     }
 
     location ~ ^/session.* {
-        proxy_pass         ${NGINX_REROUTE_HOST};
+        proxy_pass http://syncing-server-js;
     }
 
     location ~ ^/auth.* {
-        proxy_pass         ${NGINX_REROUTE_HOST};
+        proxy_pass http://syncing-server-js;
```
standarius, line 27: so `auth` locations get proxied to `syncing-server-js`: is this intended?

baptiste-grob:
Have you tried making the switch to port `3124` already? SNJS is equipped to handle the `data` and `meta` responses since #216 got merged

docker/syncing-server.env.sample
karolsojko: my bad - will fix

test/proxy.conf.template
karolsojko: so this is Load Balancer config ... which is routing traffic only between syncing-server and syncing-server-js

whereas `auth` is standing behind the API Gateway ... so only urls like `.../v1/..` go to API Gateway and land on Auth

karolsojko:
> Have you tried making the switch to port `3124` already? SNJS is equipped to handle the `data` and `meta` objects since #216 got merged

@baptiste-grob I did not - I'll try in a next PR - just wanted to get this setup in before trying that switch

baptiste-grob (approved):

standarius (approved):

PR #239, opened 2021-03-16
History refactor

mobitar:
- Simplifies history structure
- Improves conflict handling by checking if client and server previously had same value for an item before conflicting. If the client and server previously had the same value, then the pending change that was rejected from the server should in fact trump as winner, since the change must be user initiated and intentional.

[Internal ticket](https://app.asana.com/0/search/1200065820969141/1200029882307982/f)

(Branch was accidentally called search-refactor, should be called history-refactor)


mobitar:
(Optional review from @standarius or @karolsojko as the changes are somewhat dense)

packages/snjs/lib/services/history/history_manager.ts
```diff
@@ -226,24 +317,95 @@ export class SNHistoryManager extends PureService {
...
+     * determine what it should keep and what it shouldn't. So, it is possible
+     * to have a threshold of 60 but have 600 entries, if the item history deems
+     * those worth keeping.
+     *
+     * Rules:
+     * - Keep an entry if it is the oldest entry
+     * - Keep an entry if it is the latest entry
+     * - Keep an entry if it is Significant
+     * - If an entry is Significant and it is a deletion change, keep the entry before this entry.
+     */
```
baptiste-grob, line 364: This comment should probably be above the function instead of inside its body, no?

packages/snjs/lib/services/history/history_manager.ts
```diff
@@ -226,24 +317,95 @@ export class SNHistoryManager extends PureService {
...
+      return undefined;
+    }
+    return new HistoryEntry(decryptedPayload as SurePayload);
+  }
+
+  optimizeHistoryForItem(uuid: string): void {
+    /**
+     * Clean up if there are too many revisions. Note itemRevisionThreshold
+     * is the amount of revisions which above, call for an optimization. An
+     * optimization may not remove entries above this threshold. It will
```
baptiste-grob, line 354: > An optimization may not remove entries above this threshold.

This would mean that on every item change, an optimize job gets started yes? Is this not a concern with regards to performance/chugging? Or do you expect this scenario to be unlikely?

packages/snjs/lib/services/history/history_manager.ts
```diff
@@ -226,24 +317,95 @@ export class SNHistoryManager extends PureService {
...
+          if (previousEntry) {
+            keepEntries.unshift(previousEntry);
+          }
+        }
+      } else {
+        /** Don't keep, remove if in keep */
+        removeFromArray(keepEntries, entry);
+      }
+    };
+    for (let index = entries.length; index--; ) {
```
baptiste-grob, line 397: Have you considered:
```ts
for (let index = entries.length - 1; index >= 0; index--) {
}
```

packages/snjs/lib/services/history/history_manager.ts
```diff
@@ -226,24 +317,95 @@ export class SNHistoryManager extends PureService {
...
+      const entry = entries[index];
+      if (index === 0 || index === entries.length - 1) {
+        /** Keep the first and last */
+        processEntry(entry, index, true);
+      } else {
+        processEntry(entry, index, isEntrySignificant(entry));
```
baptiste-grob, line 403: You could do
```ts
const isEntrySignificant = index === 0 || index === entries.length || isEntrySignificant(entry);
```
And remove the need for a nested function altogether to do your processing inline, thus having a clean up-to-down reading flow

test/sync_tests/conflicting.test.js
```diff
@@ -0,0 +1,763 @@
...
+      PayloadSource.LocalChanged
+    );
+    this.expectedItemCount++;
+    await this.application.syncService.sync(syncOptions);
+    /** First modify the item without saving so that
+     * our local contents digress from the server's */
+    await this.application.changeItem(item.uuid, (mutator) => {
+      mutator.content.foo = `${Math.random()}`;
+    });
+    await this.application.changeAndSaveItem(
```
baptiste-grob, line 67: How does `changeAndSaveItem()` not override what `changeItem()` did? I don't understand how the conflict is created…

test/sync_tests/conflicting.test.js
```diff
@@ -0,0 +1,763 @@
...
+      PayloadSource.LocalChanged
+    );
+    this.expectedItemCount++;
+    await this.application.syncService.sync(syncOptions);
+    /** First modify the item without saving so that
+     * our local contents digress from the server's */
+    await this.application.changeItem(item.uuid, (mutator) => {
+      mutator.title = `${Math.random()}`;
+    });
+    await this.application.changeAndSaveItem(
```
baptiste-grob, line 96: Same comment as above:
> How does `changeAndSaveItem()` not override what `changeItem()` did? I don't understand how the conflict is created…

test/sync_tests/conflicting.test.js
```diff
@@ -0,0 +1,763 @@
...
+    this.expectedItemCount++;
+    const allItems = this.application.itemManager.items;
+    expect(allItems.length).to.equal(this.expectedItemCount);
+
+    const note1 = this.application.itemManager.notes[0];
+    const note2 = this.application.itemManager.notes[1];
+    expect(note1.content.title).to.not.equal(note2.content.title);
+  });
+
+  it('basic conflict with clearing local state', async function () {
```
baptiste-grob, line 249: What does this test?

test/sync_tests/conflicting.test.js
```diff
@@ -0,0 +1,763 @@
...
+      },
+      undefined,
+      undefined,
+      syncOptions
+    );
+
+    // conflict_of is a key to ignore when comparing content, so item should
+    // not be duplicated.
+    await this.application.syncService.sync(syncOptions);
+  });
```
baptiste-grob, line 368: Are you confident in using `afterEach` hooks for assertions that are core to an individual test, instead of just cleanup? It seems dangerous to me, but I've not explored this thought deeply and I trust your judgement. Just want to make sure this is your intent.

test/sync_tests/conflicting.test.js
```diff
@@ -0,0 +1,763 @@
...
+      syncOptions
+    );
+
+    // We expect that this item maintained.
+    expect(this.application.itemManager.items.length).to.equal(
+      this.expectedItemCount
+    );
+  });
+
+  it('should create conflict if syncing an item that is stale', async function () {
```
baptiste-grob, line 478: What constitutes a stale item?

test/sync_tests/conflicting.test.js
```diff
@@ -0,0 +1,763 @@
...
+    expect(this.application.itemManager.items.length).to.equal(
+      this.expectedItemCount
+    );
+  });
+
+  it('handles stale data in bulk', async function () {
+    /** This number must be greater than the pagination limit per sync request.
+     * For example if the limit per request is 150 items sent/received, this number should
+     * be something like 160. */
+    const largeItemCount = 160;
```
baptiste-grob, line 538: Any way to access this constant from the application code? As in, have something like:
```ts
const largeItemCount = this.application.syncService.largeitemCount + 1
```
This would prevent some test/code drifting

baptiste-grob:
Please feel free to not address the more stylistic comments, they are simply there for you own consideration :)

packages/snjs/lib/protocol/collection/item_collection.ts
```diff
@@ -105,7 +100,7 @@ export class ItemCollection extends MutableCollection<SNItem> {
 
   /** Returns the filtered and sorted list of elements for this content type,
    * according to the options set via `setDisplayOptions` */
-  public displayElements(contentType: ContentType) {
+  public displayElements(contentType: ContentType): Array<SNItem | undefined> {
```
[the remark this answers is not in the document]
mobitar: Yeah I'm not sure why it's returning undefined in the first place. Seems to do just fine without it.

packages/snjs/lib/services/history/entries/generator.ts
```diff
@@ -0,0 +1,28 @@
...
+}
+
+function historyClassForContentType(contentType: ContentType) {
+  switch (contentType) {
+    case ContentType.Note:
+      return NoteHistoryEntry;
+    default:
+      return HistoryEntry;
+  }
+}
```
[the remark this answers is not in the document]
mobitar: Ah nice, didn't know that. Looks like I can remove the cast with the current code though and it passes.

packages/snjs/lib/services/history/entries/history_entry.ts
```diff
@@ -0,0 +1,99 @@
...
+      lhs.userModifiedDate.getTime() === rhs.userModifiedDate.getTime();
+    if (!datesEqual) {
+      return false;
+    }
+    /** Dates are the same, but because JS is only accurate to milliseconds,
+     * items have different content but same dates */
+    return lhs.isItemContentEqualWith(rhs);
+  }
+
+  public operationVector(): number {
```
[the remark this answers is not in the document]
mobitar: I think so. A "direction vector".

packages/snjs/lib/services/history/history_manager.ts
```diff
@@ -33,189 +54,259 @@ const PERSIST_TIMEOUT = 2000;
...
+      }
+      const payload = item.payload;
+      if (
+        item.deleted ||
+        payload.format !== PayloadFormat.DecryptedBareObject
+      ) {
+        continue;
+      }
+      const itemHistory = this.history[item.uuid] || [];
+      const latestEntry = itemHistory.length > 0 ? itemHistory[0] : undefined;
```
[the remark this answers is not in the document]
mobitar: So this one was really strange.

```
const latestEntry = itemHistory.length > 0 ? itemHistory[0] : undefined;
```

Above `latestEntry` has type `HistoryEntry | undefined`, which is what I want.

```
const latestEntry = itemHistory[0];
```

And here `latestEntry` has type `HistoryEntry`, which is not what I want.

packages/snjs/lib/services/history/history_manager.ts
```diff
@@ -33,189 +54,259 @@ const PERSIST_TIMEOUT = 2000;
...
+      const stagedEntry = this.historyStaging[item.uuid];
+      /** Add prospective to staging, and consider now adding previously staged as new revision */
+      this.historyStaging[item.uuid] = currentValueEntry;
+      if (!stagedEntry) {
+        continue;
+      }
+      if (stagedEntry.isSameAsEntry(currentValueEntry)) {
+        continue;
+      }
+      if (latestEntry && stagedEntry.isSameAsEntry(latestEntry)) {
```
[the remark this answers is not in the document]
mobitar: I don't think so. `latestEntry` represents the most recent revision in the history array. And `currentValueEntry` represents the current value of the item if it were in the body of a HistoryEntry.

packages/snjs/lib/services/history/history_manager.ts
mobitar: An optimize could be called on every change. It's not too resource intensive.

test/sync_tests/conflicting.test.js
mobitar: changeItem changes the item but does not sync. changeAndSaveItem changes and syncs.

The goal is to change the value locally before we sync so that by the time conflicts are processed, our previous revision's value has a different value from the server.

test/sync_tests/conflicting.test.js
mobitar: That a conflict is created. This is via `this.expectedItemCount++;`

test/sync_tests/conflicting.test.js
mobitar: Good point

test/sync_tests/conflicting.test.js
mobitar: It's mostly this: `mutator.updated_at = Factory.yesterday()`

packages/snjs/lib/protocol/payloads/deltas/conflict.ts
```diff
@@ -11,29 +12,27 @@ import {
...
+    protected readonly source: PayloadSource,
+    protected readonly historyMap?: HistoryMap
+  ) {}
 
   public async resultingCollection(): Promise<ImmutablePayloadCollection> {
     const tmpBaseItem = CreateItemFromPayload(this.basePayload);
     const tmpApplyItem = CreateItemFromPayload(this.applyPayload);
-    const strategy = tmpBaseItem.strategyWhenConflictingWithItem(tmpApplyItem);
+    const historyEntries =
+      (this.historyMap && this.historyMap?.[this.basePayload.uuid]) || [];
```
baptiste-grob, line 30: `const historyEntries = this.historyMap?.[this.basePayload.uuid]) || []` should be sufficient since you're using `?.`

packages/snjs/lib/services/history/history_manager.ts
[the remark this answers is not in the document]
baptiste-grob: Oh wow yeah, that is really strange. Maybe it's a TypeScript default because it would be a pain in other instances?

I guess you could always do this:
```ts
 const latestEntry: HistoryEntry | undefined = itemHistory[0];
 ```
 
 But at that point it's about as busy as your original line so 🤷‍♀️

File: packages/snjs/lib/services/history/history_manager.ts (hunk 15 above)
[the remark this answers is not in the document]
Comment (baptiste-grob) [reply packages/snjs/lib/services/history/history_manager.ts:183]: Gotcha!

File: test/sync_tests/conflicting.test.js (hunk 5 above)
Comment (baptiste-grob) [reply test/sync_tests/conflicting.test.js:70]: Since changeAndSaveItem is essentially calling changeItem and then syncing, and you're doing it in the same item, I read this as changing the item twice locally and then syncing, which I wouldn't expect to create a conflict.

File: packages/snjs/lib/models/core/item.ts [hunk 17]
```diff
@@ -310,6 +326,20 @@ export class SNItem {
     ]);
     if (itemsAreDifferentExcludingRefs) {
       const twentySeconds = 20_000;
+      if (history) {
```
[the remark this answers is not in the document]
mobitar: Oh wow. Looks like it's coming from the TS standard lib.
Screen Shot 2021-03-25 at 11 56 44 AM

mobitar:
>  const latestEntry: HistoryEntry | undefined = itemHistory[0];

Surprisingly this still doesn't have the desired effect. It shows the correct type, but still allows you to use it as if it were 100% not undefined.

packages/snjs/lib/models/core/item.ts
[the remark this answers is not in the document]
standarius: it's simply https://developer.mozilla.org/en-US/docs/Web/API/Window/history
possibly this linter rule could help: https://eslint.org/docs/rules/no-restricted-globals

baptiste-grob (approved):

PR #241, opened 2021-03-23
feat: protect passcode creation

baptiste-grob:


packages/snjs/lib/application.ts
```diff
@@ -909,7 +909,8 @@ export class SNApplication {
       }
       if (hasPasscode) {
         /* Upgrade passcode version */
-        await this.changePasscode(
+        await this.removePasscodeWithoutWarning();
+        await this.setPasscodeWithoutWarning(
```
mobitar, line 913: Just to be sure: will this double prompt for the account password (because we now ask for account password to add a passcode), after being initially prompted via the challenge on line 874?

baptiste-grob: This makes sure there is no double-prompting by using the internal methods that don't run any checks, yes!

mobitar (approved):

PR #246, opened 2021-04-01
localStorage repair

mobitar:
- creates credential service to centralize account and passcode operations
- fixes issue where if localStorage is wiped by browser improperly (but IndexedDB remains), we perform a corrective sign in that doesn't mark all existing items as dirty before syncing.
- passes items key into transient decryption on sync response
- add traps to avoid duplicating errored payloads

### DIFF
--- a/packages/snjs/lib/services/credential_service.ts
+++ b/packages/snjs/lib/services/credential_service.ts
@@ -1,3 +1,4 @@
+import { SNProtectionService } from './protection_service';
 import { Uuid } from '@Lib/uuid';
 import { isNullOrUndefined } from '@Lib/utils';
 import {
@@ -58,7 +59,8 @@ export class SNCredentialService extends PureService<AccountEvent> {
     private itemManager: ItemManager,
     private protocolService: SNProtocolService,
     private alertService: SNAlertService,
-    private challengeService: ChallengeService
+    private challengeService: ChallengeService,
+    private protectionService: SNProtectionService
   ) {
     super();
   }
@@ -72,6 +74,7 @@ export class SNCredentialService extends PureService<AccountEvent> {
     (this.protocolService as unknown) = undefined;
     (this.alertService as unknown) = undefined;
     (this.challengeService as unknown) = undefined;
+    (this.protectionService as unknown) = undefined;
   }
 
   /**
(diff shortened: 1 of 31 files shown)


test/auth-fringe-cases.test.js
```diff
@@ -0,0 +1,144 @@
...
+        context.password,
+        undefined,
+        undefined,
+        undefined,
+        awaitSync
+      );
+      const refreshedNote = restartedApplication.itemManager.findItem(
+        note.uuid
+      );
+      expect(refreshedNote.errorDecrypting).to.equal(false);
```
baptiste-grob, line 88: Don't you also want to check that there aren't duplicate notes?

packages/snjs/lib/services/sync/sync_service.ts
```diff
@@ -586,6 +591,9 @@ export class SNSyncService extends PureService<
         throw Error(`Unhandled timing strategy ${useStrategy}`);
       }
     }
+    if (this.dealloced) {
+      return;
+    }
```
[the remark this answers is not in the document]
mobitar: Yeah, not something I wanted in either tbh. But basically this can happen if on sync completion the sync function wants to sync again, and so begins syncing, but then in a test somewhere, the application is destroyed.

packages/snjs/lib/services/credential_service.ts
```diff
@@ -0,0 +1,513 @@
...
+  public async correctiveSignIn(rootKey: SNRootKey): Promise<SignInResponse> {
+    this.lockSyncing();
+    const response = await this.sessionManager.bypassChecksAndSignInWithRootKey(
+      rootKey.keyParams.identifier,
+      rootKey
+    );
+    if (!response.error) {
+      await this.notifyEvent(AccountEvent.SignedInOrRegistered);
+      this.unlockSyncing();
+      this.syncService.downloadFirstSync(1_000, {
```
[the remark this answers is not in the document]
mobitar: Yeah not triggering any warnings. Not sure why.

packages/snjs/lib/services/singleton_manager.ts
```diff
@@ -173,14 +173,7 @@ export class SNSingletonManager extends PureService {
...
     if (handled.length > 0 && eventSource === SyncEvent.FullSyncCompleted) {
-      /**
-       * Do not await. We want any local-side changes to
-       * be awaited but the actual sync shouldn't be since it's non-essential
-       * Perform after timeout so that we can yield to event notifier that triggered us
-       */
-      setTimeout(() => {
-        this.syncService?.sync();
-      });
+      await this.syncService?.sync();
```
baptiste-grob, line 176: I assume the situation has changed?

baptiste-grob (approved):
Awesome 🎉

packages/snjs/lib/services/singleton_manager.ts
mobitar: It was difficult to create tests where you can be sure that there is no background sync happening without awaiting here. And I think in general all sync calls should be awaitable and not detached. And no test failed by awaiting here 🤷‍♀️

baptiste-grob (approved):

PR #249, opened 2021-04-05
feat: add linter rules: varsIgnorePattern, curly/all

standarius:
As discussed with @mobitar and @karolsojko, added rules for ignoring vars (`_name`) and to require curly braces around if/else at all times.

### DIFF
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@standardnotes/config",
-  "version": "1.4.0",
+  "version": "2.0.0",
   "engines": {
     "node": ">=14.0.0 <16.0.0"
   },
--- a/packages/config/package.json
+++ b/packages/config/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@standardnotes/config",
-  "version": "1.3.0",
+  "version": "1.4.0",
   "engines": {
     "node": ">=14.0.0 <16.0.0"
   },
--- a/packages/config/src/.eslintrc
+++ b/packages/config/src/.eslintrc
@@ -17,7 +17,8 @@
         "vars": "all",
         "args": "after-used",
         "ignoreRestSiblings": false,
-        "argsIgnorePattern": "^_"
+        "argsIgnorePattern": "^_",
+        "varsIgnorePattern": "^_"
       }
     ],
     "@typescript-eslint/no-explicit-any": ["error", { "ignoreRestArgs": true }],
@@ -26,6 +27,7 @@
     "quotes": ["error", "single"],
     "object-curly-spacing": ["error", "always"],
     "indent": ["error", 2],
-    "comma-dangle": ["error", "always-multiline"]
+    "comma-dangle": ["error", "always-multiline"],
+    "curly": ["error", "all"]
   }
 }
--- a/packages/config/src/.prettierrc
+++ b/packages/config/src/.prettierrc
@@ -1,3 +1,4 @@
 {
-  "singleQuote": true
+  "singleQuote": true,
+  "trailingComma": "all"
 }


mobitar (approved):

daianaszwimer (approved):

standarius:
Added more people as reviewers, as this will potentially affect all dependents of `config` which I presume are many. If anybody has a strong objection to any of the rules, please let me know.

FYI: I am in favor of relaxing `curly` to `multi-line` (https://eslint.org/docs/rules/curly#multi-line), as it does save some lines, especially when doing multiple consecutive if-return statements (factor of 3), but this is minor.

karolsojko (approved):

packages/config/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@standardnotes/config",
-  "version": "1.3.0",
+  "version": "1.4.0",
```
baptiste-grob, line 3:
```suggestion
  "version": "2.0.0",
```
This would break consumer projects which have incompatible code so I would suggest incrementing the major version number to 2

packages/config/src/.eslintrc
```diff
@@ -26,6 +27,7 @@
     "quotes": ["error", "single"],
     "object-curly-spacing": ["error", "always"],
     "indent": ["error", 2],
-    "comma-dangle": ["error", "always-multiline"]
+    "comma-dangle": ["error", "always-multiline"],
```
baptiste-grob, line 30: Prettier removes dangling commas from function declarations/calls by default, so `.prettierrc` would need to be updated as well

baptiste-grob (changes requested):

packages/config/package.json
karolsojko: consumer projects should be locked on fixed version to avoid situations like this. `2.0.0` is a major change - more of a project refactor kind of thing

packages/config/src/.eslintrc
standarius: ok, will update that

this rule was not added in this PR though ;)
I only added the last comma in it, to add the next line
that's why it's highlighted
which is a meta-illustration of why it's a good rule diff-wise.
if there was a trailing comma, the line wouldn't be confusingly highlighted

unfortunately trailing commas in JSON are not possible, so it doesn't apply here (although this seems to be JSONC which apparently can tolerate trailing commas)

packages/config/package.json
standarius: so a question of semver interpretation: does this qualify as a major change?
semver says "increment the MAJOR version when you make incompatible API changes"
no API here, so do we treat a breaking change as an incompatible API change?
or do we differentiate (as semver seems to) between major breaking changes and non-major and only increment on major?
if so -- which is this one?
I don't know, it's all the same to me, but I prefer to stay on the safe side, so I am willing to bump it up to 2.0.0.
@karolsojko ok/nok?

baptiste-grob: If you update your dependency to another minor version, I personally wouldn't expect my build to stop working, which is what will happen with this change, hence my suggestion to increment the major version number. 
Though I also understand this is not much of a concern since it can easily be fixed :)

packages/config/src/.eslintrc
baptiste-grob: Hah, my bad 😄 that's a good demonstration of the problem indeed

baptiste-grob (approved):

daianaszwimer (approved):

packages/config/package.json
karolsojko: well, we're installing with the lockfile in mind so updating a version can cause breaks. The specifics of `config` package is a bit different to be honest - every change is something that you need to adjust in your code to be honest - but then again you're upgrading the package intentionally.

karolsojko: Having that said I'm against `2.0.0` since we would very quickly reach very odd package version like `54.0.0`

standarius: As I said, I don't have a strong opinion on this, but @baptiste-grob's heuristic (if the build in dependents will break, increment the major version) does sound convincing.

On the other hand, the build potentially breaks when adding any rule, if a dependent doesn't already follow it, so as @karolsojko points out, we would be incrementing the major version quite often.

High major versions might be fine though.

All in all, this doesn't seem very important, so to avoid prolonged bikeshedding I will leave the decision to the first of:

1. Whoever has overruling privilege here -- @karolsojko?, @mobitar?. Please comment if you would like to overrule.

2. Popular vote: :+1:  this post if you think the version should be '2.0.0', :-1: this post if you think it should be '1.4.0'.

Will merge tomorrow.

karolsojko: I'm giving the choice to vox populi as this is something we all will use :)

standarius: No strong opposition to 2.0, wins by votes, no PR update required, so merging. ;)
The debate can always continue at the next opportunity.

PR #254, opened 2021-04-15
Feature: use settings API for MFA

standarius:
Related tasks:
* (obsolete) [[0]](https://app.asana.com/0/1199914526147392/1200004328990530/f)
* [[1]](https://app.asana.com/0/1185703377033953/1199671860849277/f) 
* [[2]](https://app.asana.com/0/1199914526147392/1200322309691511/f)

### DIFF
--- a/packages/snjs/lib/application.ts
+++ b/packages/snjs/lib/application.ts
@@ -101,6 +101,7 @@ import { PayloadFormat } from './protocol/payloads';
 import { SNPermissionsService } from './services/permissions_service';
 import { ProtectionEvent } from './services/protection_service';
 import { Permission } from '@standardnotes/auth';
+import { SNMfaService } from './services/api/mfa_service';
 
 /** How often to automatically sync, in milliseconds */
 const DEFAULT_AUTO_SYNC_INTERVAL = 30_000;
@@ -142,6 +143,7 @@ export class SNApplication {
   private preferencesService!: SNPreferencesService;
   private permissionsService!: SNPermissionsService;
   private credentialService!: SNCredentialService;
+  private mfaService!: SNMfaService;
 
   private eventHandlers: ApplicationObserver[] = [];
   private services: PureService<any, any>[] = [];
@@ -1395,6 +1397,7 @@ export class SNApplication {
     this.storageService.encryptionDelegate = encryptionDelegate;
     this.createChallengeService();
     this.createHttpManager();
+    this.createMfaService();
     this.createApiService();
     this.createSessionManager();
     this.createMigrationService();
@@ -1430,6 +1433,7 @@ export class SNApplication {
     (this.preferencesService as unknown) = undefined;
     (this.permissionsService as unknown) = undefined;
     (this.credentialService as unknown) = undefined;
+    (this.mfaService as unknown) = undefined;
 
     this.services = [];
   }
@@ -1477,7 +1481,9 @@ export class SNApplication {
       this.httpService,
       this.storageService,
       this.permissionsService,
-      this.defaultHost
+      this.defaultHost,
+      this.mfaService,
+      this.alertService,
     );
     this.services.push(this.apiService);
   }
@@ -1566,7 +1572,7 @@ export class SNApplication {
       this.apiService,
       this.alertService,
       this.protocolService,
-      this.challengeService
+      this.challengeService,
     );
     this.serviceObservers.push(
       this.sessionManager.addEventObserver(async (event) => {
@@ -1695,6 +1701,11 @@ export class SNApplication {
       })
     );
     this.services.push(this.preferencesService);
+  }
+
+  private createMfaService() {
+    this.mfaService = new SNMfaService(this.challengeService);
+    this.services.push(this.mfaService);
   }
 
   private getClass<T>(base: T) {
--- a/packages/snjs/lib/services/api/api_service.ts
+++ b/packages/snjs/lib/services/api/api_service.ts
@@ -68,7 +68,7 @@ type PathsV1 = {
   }) => string;
 };
 
-type DisableMfaDto = {
+type DeleteSettingDto = {
   userUuid: Uuid;
   settingName: string;
 };
@@ -77,7 +77,7 @@ type SettingProps = {
   value: string;
   serverEncryptionVersion?: number;
 };
-type EnableMfaDto = {
+type PutSettingDto = {
   userUuid: Uuid;
   props: SettingProps;
 };
@@ -104,18 +104,16 @@ const Paths: {
   // todo: add more
   v1: {
     authMethods: '/v1/auth/methods',
-    // todo: use
     keyParams: '/v1/login-params',
-    // todo: use
     register: '/v1/users',
     signIn: '/v1/login',
     signOut: '/v1/logout',
+    putSetting: (userUuid: Uuid) => `/v1/users/${userUuid}/settings`,
+    deleteSetting: ({ userUuid, settingName }: DeleteSettingDto) =>
+      `/v1/users/${userUuid}/settings/${settingName}`,
     // note: not sure about what userId is
     // todo: use
     changePassword: (userId: string) => `/v1/users/${userId}/password`,
-    putSetting: (userUuid: Uuid) => `/v1/users/${userUuid}/settings`,
-    deleteSetting: ({ userUuid, settingName }: DisableMfaDto) =>
-      `/v1/users/${userUuid}/settings/${settingName}`,
   },
 };
 
@@ -330,7 +328,7 @@ export class SNApiService extends PureService {
   public enableMfa({
     userUuid,
     props,
-  }: EnableMfaDto): Promise<HttpResponse<unknown>> {
+  }: PutSettingDto): Promise<HttpResponse<unknown>> {
     return this.request({
       verb: HttpVerb.Put,
       url: joinPaths(this.host, Paths.v1.putSetting(userUuid)),
@@ -340,7 +338,7 @@ export class SNApiService extends PureService {
     });
   }
 
-  public disableMfa(dto: DisableMfaDto): Promise<HttpResponse<unknown>> {
+  public disableMfa(dto: DeleteSettingDto): Promise<HttpResponse<unknown>> {
     return this.request({
       verb: HttpVerb.Put,
       url: joinPaths(this.host, Paths.v1.deleteSetting(dto)),
@@ -357,7 +355,7 @@ export class SNApiService extends PureService {
     return this.requestWithMfa({
       identifier: email,
       verb: HttpVerb.Get,
-      url: joinPaths(this.host, Paths.v0.keyParams),
+      url: joinPaths(this.host, Paths.v1.keyParams),
       fallbackErrorMessage: messages.API_MESSAGE_GENERIC_INVALID_LOGIN,
       params,
       /** A session is optional here, if valid, endpoint returns extra params */
@@ -377,7 +375,7 @@ export class SNApiService extends PureService {
       ) as RegistrationResponse;
     }
     this.registering = true;
-    const url = joinPaths(this.host, Paths.v0.register);
+    const url = joinPaths(this.host, Paths.v1.register);
     const params = this.params({
       password: serverPassword,
       email,
--- a/packages/snjs/lib/services/api/api_service.ts
+++ b/packages/snjs/lib/services/api/api_service.ts
@@ -19,7 +19,6 @@ import { PurePayload } from '@Payloads/pure_payload';
 import { SNRootKeyParams } from './../../protocol/key_params';
 import { SNStorageService } from './../storage_service';
 import {
-  ErrorTag,
   HttpParams,
   HttpRequest,
   HttpVerb,
@@ -32,8 +31,13 @@ import { PureService } from '@Services/pure_service';
 import { joinPaths } from '@Lib/utils';
 import { StorageKey } from '@Lib/storage_keys';
 import { SNPermissionsService } from '../permissions_service';
+import { GetAuthMethodsResponse, MfaStatus, SNMfaService } from './mfa_service';
+import { SNAlertService } from '@Lib/index';
+import { ErrorTag } from '@standardnotes/auth';
+import { Uuid } from '@standardnotes/auth';
 
-type PathNames = {
+type PathsV0 = {
+  authMethods: string;
   keyParams: string;
   register: string;
   signIn: string;
@@ -47,10 +51,43 @@ type PathNames = {
   itemRevision: (itemId: string, revisionId: string) => string;
 };
 
+type PathsV1 = {
+  authMethods: string;
+  keyParams: string;
+  register: string;
+  signIn: string;
+  signOut: string;
+  changePassword: (userId: string) => string;
+  putSetting: (userUuid: Uuid) => string;
+  deleteSetting: ({
+    userUuid,
+    settingName,
+  }: {
+    userUuid: Uuid;
+    settingName: string;
+  }) => string;
+};
+
+type DisableMfaDto = {
+  userUuid: Uuid;
+  settingName: string;
+};
+type SettingProps = {
+  name: string;
+  value: string;
+  serverEncryptionVersion?: number;
+};
+type EnableMfaDto = {
+  userUuid: Uuid;
+  props: SettingProps;
+};
+
 const Paths: {
-  v0: PathNames;
+  v0: PathsV0;
+  v1: PathsV1;
 } = {
   v0: {
+    authMethods: '/auth/methods',
     keyParams: '/auth/params',
     register: '/auth',
     signIn: '/auth/sign_in',
@@ -64,12 +101,32 @@ const Paths: {
     itemRevision: (itemId: string, revisionId: string) =>
       `/items/${itemId}/revisions/${revisionId}`,
   },
+  // todo: add more
+  v1: {
+    authMethods: '/v1/auth/methods',
+    // todo: use
+    keyParams: '/v1/login-params',
+    // todo: use
+    register: '/v1/users',
+    signIn: '/v1/login',
+    signOut: '/v1/logout',
+    // note: not sure about what userId is
+    // todo: use
+    changePassword: (userId: string) => `/v1/users/${userId}/password`,
+    putSetting: (userUuid: Uuid) => `/v1/users/${userUuid}/settings`,
+    deleteSetting: ({ userUuid, settingName }: DisableMfaDto) =>
+      `/v1/users/${userUuid}/settings/${settingName}`,
+  },
 };
 
 /** Legacy api version field to be specified in params when calling v0 APIs. */
 const V0_API_VERSION = '20200115';
 
 type InvalidSessionObserver = (revoked: boolean) => void;
+
+type HttpRequestWithFallbackError = HttpRequest & {
+  fallbackErrorMessage: string;
+};
 
 export class SNApiService extends PureService {
   private session?: Session;
@@ -85,7 +142,9 @@ export class SNApiService extends PureService {
     private httpService: SNHttpService,
     private storageService: SNStorageService,
     private permissionsService: SNPermissionsService,
-    private host: string
+    private host: string,
+    private mfaService: SNMfaService,
+    private alertService: SNAlertService,
   ) {
     super();
   }
@@ -188,13 +247,9 @@ export class SNApiService extends PureService {
     }
   }
 
-  private async request(params: {
-    verb: HttpVerb;
-    url: string;
-    fallbackErrorMessage: string;
-    params?: HttpParams;
-    authentication?: string;
-  }) {
+  private async request(
+    params: HttpRequestWithFallbackError,
+  ): Promise<HttpResponse<unknown>> {
     try {
       const response = await this.httpService.runHttp(params);
       this.processResponse(response);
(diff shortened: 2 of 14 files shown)


packages/snjs/lib/services/api/mfa_service.ts
```diff
@@ -0,0 +1,169 @@
...
+  status: Exclude<MfaStatus, 
+    MfaStatus.EnteredCode | 
+    MfaStatus.ErrorResponse
+  >,
+}
+
+
+// todo: figure out if and where this belongs
+// possibly this belongs in the Auth SDK and it should be shared with Auth
+type GetAuthMethodsResponse = {
```
standarius, line 43: This is the type of the response from the new /auth/methods endpoint copied from Auth. It's a part of a common interface which seems to live in the auth package (Auth SDK) in SNJS -- so it most likely belongs there.

packages/snjs/lib/services/api/mfa_service.ts
```diff
@@ -0,0 +1,169 @@
...
+export class SnMfaService extends PureService {
+  constructor(
+    private apiService: SNApiService,
+    private alertService: SNAlertService,
+    private challengeService: ChallengeService,
+  ) {
+    super()
+  }
+
+  async handleMfaOld({
```
standarius, line 64: I wanted to first refactor the old MFA handling, test that, and then swap the implementation for the new one, but this turns out to be quite involved, so I'll drop this and implement the new directly.

packages/snjs/lib/application.ts
```diff
@@ -145,6 +146,7 @@ export class SNApplication {
   private keyRecoveryService!: SNKeyRecoveryService;
   private preferencesService!: SNPreferencesService;
   private permissionsService!: SNPermissionsService;
+  private mfaService!: SnMfaService;
```
mobitar, line 149:
```suggestion
  private mfaService!: SNMfaService;
```

Would go for consistency for now, until we decide to rename all classes to proper pascal case if that's what we want to do.

packages/snjs/lib/services/api/api_service.ts
```diff
@@ -234,6 +236,21 @@ export class SNApiService extends PureService {
     });
   }
 
+  async getAuthMethods({
+    email,
```
mobitar, line 240: I would again aim for consistency here rather than creating a hybrid API of DTO and non-DTO. Personally TypeScript has removed the need for DTOs for me. Might make sense for function with an excessive amount of arguments but here seems unnecessary.

packages/snjs/lib/services/api/mfa_service.ts
```diff
@@ -0,0 +1,169 @@
...
+export enum MfaStatus {
+  EnteredCode = 'mfa-entered-code',
+  NotRequired = 'mfa-not-required',
+  Ignored = 'mfa-ignored',
+
+  ErrorMissingCode = 'mfa-error-missing-code',
+  ErrorResponse = 'mfa-error-response',
+}
+
+export type MfaResultOld = {
```
mobitar, line 17: I would quantify what "Old" means. Either in terms of a specific API version, or a date, or perhaps type of server.

mobitar: (Not in terms of a comment btw, but as the name of the type itself)

packages/snjs/lib/services/api/mfa_service.ts
```diff
@@ -0,0 +1,169 @@
...
+  mfaKey: string,
+  enteredMfaCode: string,
+} | {
+  status: Exclude<MfaStatus, MfaStatus.EnteredCode>,
+  response: HttpResponse<unknown>,
+}
+
+export type MfaResult = {
+  status: MfaStatus.EnteredCode,
+  totpUuid: string,
```
mobitar, line 28: Wouldn't this be uuid of the UserSetting and not the totp?

packages/snjs/lib/services/api/mfa_service.ts
mobitar, line 64:
```suggestion
 private async handleMfaOld({
```

Add access modifier.

packages/snjs/lib/application.ts
standarius: I normally go by the Google's style guide on this[1](https://google.github.io/styleguide/javaguide.html#s5.3-camel-case), [2](https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined), because it's the most reasonable and regular take on this that I know of.

But I'm ok with whatever is satisfactory to everyone else as long as there are some rules behind it that can be followed. 

So the rule I can detect in SNJS is to always capitalize SN, otherwise follow regular camelCase. Is this correct?

Outside of SNJS it definitely isn't (e.g. there are such names as VerifyMFA, VerifyMFADTO, HTTPClient) -- going by these the name would be `SNMFAService`, but this is not very readable indeed.

PS on that topic, the Web API built-in `XMLHttpRequest` is a nice illustration of camelCase/PascalCase being confusing -- that's probably why the Google guides have it as an example. :D

packages/snjs/lib/services/api/api_service.ts
standarius: I don't really see this as DTO (which seems to be a pretty vague concept), but as named arguments (which aren't really supported in JS). They are the better default, even if you have a super-strong static type system, because such system will not help you with positional arguments of the same type next to each other. Hence my habit of always starting with that in JS/TS. Got bitten by this enough times that I like to err on the safe side.

And named arguments are less brittle in terms of cosmetic refactoring -- you can rearrange all you want without fear of breaking something.

For a single argument this is indeed not necessary and I would probably change this before submitting this PR for review.

packages/snjs/lib/services/api/mfa_service.ts
standarius: This shall be removed before undrafting this PR.

packages/snjs/lib/services/api/mfa_service.ts
standarius: Good catch, will fix.

packages/snjs/lib/services/api/mfa_service.ts
standarius: This shall be removed before undrafting this PR. See https://github.com/standardnotes/snjs/pull/254#discussion_r614263149

Meant to be public though.

packages/snjs/lib/services/api/api_service.ts
mobitar: How do you specify if an argument is required or optional with named args? Maybe all required args can be statically typed, and the rest can be via "DTO" or named args?

packages/snjs/lib/application.ts
mobitar: Correct, the rule is capital SN, then PascalCase for the rest. The reason for the prefix btw is that each of the services can be subclassed by clients, so it makes it easier to detect what is the root class and what is the subclass. So for example mobile will subclass SNComponentManager as ComponentManager. Whereas if base class was ComponentManager the clients might have to do funky naming to avoid conflicting. (Or maybe they could just do WebComponentManager I suppose..)

mobitar: Also it was meant to allow clients to quickly see whether something came from SNJS or from the client code. But ultimately perhaps not necessary. So I'm in favor of generally refactoring this in bulk, but until we do, to have it blend in with the present style.

packages/snjs/lib/services/api/api_service.ts
standarius: > How do you specify if an argument is required or optional with named args?

Similarly to positional:

```
{
  requiredArg,
  optionalArg = default,
}: {
  requiredArg: Type1
  optionalArg?: Type2,
}
```

> Maybe all required args can be statically typed, and the rest can be via "DTO" or named args?

All should be statically typed unless there is a really good reason not to.

If you mean required should be positional and optional should be named -- yes, that often happens, but it doesn't solve the issues of mixing up order for args of the same type and impractical refactorability -- they are orthogonal to that rule.

***

The major objection to named args is verbosity, especially in TS -- you can't type each arg inline. 

Hairs can also be split in JS about performance.

Other than that, named are just better for humans and you can't make a positional error with them.

That said, the objections are sometimes valid and reason has to be applied. No need to stick named args everywhere if it may improve things.

2 or 3 positional args might be fine, so long as they are of different types. Same types could also be ok if the order doesn't really matter (if the operation is commutative and guaranteed to stay that way -- but again, you have to think this through case-by-case).

packages/snjs/lib/application.ts
standarius: Uppercased the N, so resolving.

packages/snjs/lib/services/api/mfa_service.ts
standarius: removed, so resolving

packages/snjs/lib/services/api/mfa_service.ts
standarius: removed, so resolving

packages/snjs/lib/services/api/api_service.ts
standarius: fixed, so resolving

packages/snjs/lib/services/api/mfa_service.ts
standarius: dropped, so resolving

packages/snjs/lib/services/api/api_service.ts
```diff
@@ -207,31 +212,80 @@ export class SNApiService extends PureService {
...
+    }) as Promise<HttpResponse<GetAuthMethodsResponse>>
+  }
+
+  private async requestWithMfa(
+    httpRequest: HttpRequestWithFallbackError & {
+      identifier: string,
+    },
+  ): Promise<HttpResponse> {
+    const {identifier, ...restRequest} = httpRequest
+    for (;;) {
```
mobitar, line 234: What is this wizardry 🧐

packages/snjs/lib/services/api/api_service.ts
```diff
@@ -207,31 +212,80 @@ export class SNApiService extends PureService {
...
+          const response = await this.request({
+            ...restRequest,
+            params: {
+              ...restRequest.params,
+              ...mfaParams,
+            }
+          })
+        
+          if (response.error?.tag === ErrorTag.InvalidMfaCode) {
+            await this.alertService.alert(messages.SignInStrings.IncorrectMfa)
```
mobitar, line 264: Not sure if this needs to be awaited. I thinking awaiting would wait on the user to dismiss the alert before continuing code execution (unless that's what you want?)

packages/snjs/lib/services/api/messages.ts
```diff
@@ -99,6 +99,7 @@ export const SignInStrings = {
     'Your passcode is required in order to sign in to your account.',
   IncorrectMfa: 'Incorrect two-factor authentication code. Please try again.',
   SignInCanceledMissingMfa: 'Your sign in request has been canceled.',
+  ErrorFetchingAuthMethods: 'There was an error when fetching authenitcation methods. Please try again.',
```
mobitar, line 102:
```suggestion
  ErrorFetchingAuthMethods: 'An error occurred while trying to fetch authentication methods for your account. Please try again.',
```

packages/snjs/lib/services/api/mfa_service.ts
```diff
@@ -0,0 +1,126 @@
...
+        const {mfaSecretUuid} = totp
+
+        const enteredMfaCode = await this.promptForMfaValue()
+
+        if (enteredMfaCode === undefined) {
+          return {
+            status: MfaStatus.ErrorMissingCode,
+          }
+        }
+
```
mobitar, line 69: What's the guideline on whitespace? When should one put whitespace and not put whitespace between lines?

packages/snjs/lib/services/api/api_service.ts
standarius: An infinite loop. Linter does not accept `while (true)`.

The loop was there previously, except implemented via recursion. It handles retrying if MFA code is incorrect.

packages/snjs/lib/services/api/api_service.ts
standarius: Same thing was happening previously:

```js
      if (signInResponse.error.payload?.mfa_key) {
        if (mfaCode) {
          await this.alertService.alert(SignInStrings.IncorrectMfa);
        }
```

`if (mfaCode)` would be true only on retry (so not the first time there was an error response) -- so the logic here is the same, except now it's more explicit.

I assume this is the intended behavior. If not -- let me know and I can change it to not await.

packages/snjs/lib/services/api/mfa_service.ts
standarius: Here I'm conforming to @karolsojko's suggestion from some time ago to put an empty line before `return`. Makes no difference to me.

Usual guidelines are to separate logically related sections with one blank line[[1]](https://google.github.io/styleguide/javaguide.html#s4.6.1-vertical-whitespace)[[2]](https://www.python.org/dev/peps/pep-0008/#blank-lines).

This method had some blank lines leftover from refactoring (55, 61) which shall be deleted in the next update.

standarius:
This is ready for review, but it depends on PR #256 which [conflicts](https://github.com/standardnotes/snjs/pull/256/files#r619071154) with #261, so I'll undraft after resolving that.

mobitar:
What's the approach with hybrid JS styles? Should we begin moving towards the future now (omitting semicolons, adding newines, etc), or only after we officially incorporate those rules in the linter? session_manager for example has both lines with semicolons and without.

karolsojko:
If we decide upon any rule we have to put it into `@standardnotes/config` and the repository utilizing it will have to abide then

standarius:
> What's the approach with hybrid JS styles? Should we begin moving towards the future now (omitting semicolons, adding newines, etc), or only after we officially incorporate those rules in the linter? session_manager for example has both lines with semicolons and without.

I habitually write without semicolons and the linter wasn't complaining, so that's the result.

For now I would allow whatever the linter is ok with, but I think we should go towards being consistent with `config` (so no semicolons) in the future -- if there is no task to do that and adjust the style, I'll make one.

I don't think there are any newline rules in the config, but we can introduce some (if there are options for that).

mobitar:
Can you also run Prettier on all the files? I'm getting huge style discrepancies on mfa_manager for example.

mobitar:
(Prettier would also add the missing semicolons according to the standards we've set)

packages/snjs/lib/services/api/api_service.ts
```diff
@@ -188,13 +198,8 @@ export class SNApiService extends PureService {
...
 
-  private async request(params: {
-    verb: HttpVerb;
-    url: string;
-    fallbackErrorMessage: string;
-    params?: HttpParams;
-    authentication?: string;
-  }) {
+  private async request(params: HttpRequestWithFallbackError):
+  Promise<HttpResponse<unknown>> {
```
baptiste-grob, line 202: Do you think it's better to have the response content be `unknown` than, say, a generic type?
```suggestion
  private async request<T>(params: HttpRequestWithFallbackError):
  Promise<HttpResponse<T>> {
```

On one hand I think `unknown` is more correct, on the other casting everywhere always seemed a little dangerous to me. Though that may not be a rational feeling.

standarius: Definitely `unknown` should be avoided as much as possible. Casting should be avoided even more. That's a sensible feeling.

Used unknown here to be consistent with what `this.httpService.runHttp` and `this.errorResponseWithFallbackMessage` return.

Fixing that will require quite a bit of refactoring (I actually started it and then backed out recognizing the extent) -- outside of scope of this PR.

I did create a task for this however: https://app.asana.com/0/1185703377033953/1200169272405083/f

baptiste-grob: Cool, thank you for expanding your reasoning for me 🙂

standarius:
This PR is also related to tickets [[1]](https://app.asana.com/0/1185703377033953/1199671860849277/f) and [[2]](https://app.asana.com/0/1199914526147392/1200322309691511/f).

standarius:
The PR depends on https://github.com/standardnotes/auth/pull/26 and dev versions of auth and api-gateway

- [ ] Release new auth and api-gateway images and update docker-compose.yml

standarius:
[ ] Make all SNJS tests pass -- probably will require adjustments related to v1 endpoints

test/mfa.test.js
```diff
@@ -0,0 +1,145 @@
...
+    const keyParamsAfterSignIn = await application.apiService.getAccountKeyParams(this.email)
+    expect(keyParamsAfterSignIn.version).to.equal(expectedVersion)
+
+    const initialEnableWhileSignedInResponse = await application.apiService.enableMfa({
+      userUuid,
+      secret,
+    })
+    expect(initialEnableWhileSignedInResponse.status).to.equal(201)
+
+    // todo: should this request be made w/o prompting for mfa code?
```
mobitar, line 100: Needs cleanup?

test/mfa.test.js
```diff
@@ -0,0 +1,145 @@
...
+    this.password = Uuid.GenerateUuidSynchronously();
+  });
+
+  afterEach(function () {
+    this.application.deinit();
+    this.application = null;
+    localStorage.clear();
+  });
+
+  it('mfa flow', async function () {
```
mobitar, line 21: This test is too long. One test is also not enough for a feature like this. Should be broken up into different tests. One test should be "removing 2FA should allow sign in without token" for example.

packages/snjs/lib/services/api/session_manager.ts
```diff
@@ -508,29 +442,8 @@ export class SNSessionManager extends PureService<SessionEvent> {
...
-          email,
-          rootKey,
-          signInResponse.error.payload.mfa_key,
-          inputtedCode
-        );
-      } else {
-        /** Some other error, return to caller */
-        return signInResponse;
-      }
+      // return error to caller
```
mobitar, line 445: Please follow SNJS comment convention: `/** Sentence case */`

packages/snjs/lib/services/api/api_service.ts
```diff
@@ -207,34 +250,127 @@ export class SNApiService extends PureService {
...
+          const response = await this.request({
+            ...restRequest,
+            params: {
+              ...restRequest.params,
+              ...mfaParams,
+            },
+          });
+
+          if (response.error?.tag === ErrorTag.MfaInvalid) {
+            // alert, then iterate (this is an infinite loop)
```
mobitar, line 308: Comment convention

test/mfa.test.js
mobitar: Should have separate tests for enabling and disabling 2FA as well, that just ensures you are prompted or not prompted.

packages/snjs/lib/services/api/api_service.ts
```diff
@@ -207,34 +250,127 @@ export class SNApiService extends PureService {
...
+  }): Promise<HttpResponse<unknown>> {
+    return this.request({
+      verb: HttpVerb.Put,
+      url: joinPaths(this.host, Paths.v1.putSetting(userUuid)),
+      fallbackErrorMessage: messages.MFA_ENABLE_ERROR,
+      authentication: this.session?.authorizationValue,
+      params: {
+        name: MfaSetting.MfaSecret,
+        value: secret,
+        serverEncryptionVersion: 1,
```
mobitar, line 333: `1` should be extracted into named const.

packages/snjs/lib/services/api/http_service.ts
```diff
@@ -99,9 +96,11 @@ export class SNHttpService extends PureService {
         this.stateChangeHandlerForRequest(request, resolve, reject);
       };
       if (
-        verb === HttpVerb.Post ||
-        verb === HttpVerb.Patch ||
-        verb === HttpVerb.Delete
+        params !== undefined &&
+        (verb === HttpVerb.Post ||
```
mobitar, line 100: Should use Array.includes instead

packages/snjs/lib/services/api/messages.ts
```diff
@@ -62,6 +62,14 @@ export const DO_NOT_CLOSE_APPLICATION =
 
 export const UNKNOWN_ERROR = 'Unknown error.';
 
+export const MFA_ENABLE_ERROR =
+  'An error occured when trying to enable multi-factor authentication. Please try again.';
```
mobitar, line 66:
```suggestion
  'An error occurred while trying to enable multi-factor authentication. Please try again.';
```

packages/snjs/lib/services/api/messages.ts
```diff
@@ -62,6 +62,14 @@ export const DO_NOT_CLOSE_APPLICATION =
...
+export const MFA_DISABLE_ERROR =
+  'An error occured when trying to disable multi-factor authentication. Please try again.';
```
mobitar, line 68:
```suggestion
  'An error occurred while trying to disable multi-factor authentication. Please try again.';
```

packages/snjs/lib/services/api/messages.ts
```diff
@@ -62,6 +62,14 @@ export const DO_NOT_CLOSE_APPLICATION =
...
+
+export const AUTH_METHODS_ERROR =
+  'An error occured when trying to fetch authentication methods. Please try again.';
```
mobitar, line 71:
```suggestion
  'An error occurred while trying to fetch authentication methods. Please try again.';
```

test/mfa.test.js
standarius: - [ ] cleanup

packages/snjs/lib/services/api/session_manager.ts
standarius: Double asterisk or triple slash comments are conventionally used for documentation. E.g. [[1]](https://en.wikipedia.org/wiki/Javadoc#Structure_of_a_Javadoc_comment), [[2]](https://jsdoc.app/about-getting-started.html).

Are we consciously breaking that convention?

packages/snjs/lib/services/api/api_service.ts
standarius: - [ ] extract

packages/snjs/lib/services/api/http_service.ts
standarius: - [ ] use Array.includes

karolsojko:
Outdated, closing.

PR #256, opened 2021-04-16
feat: proxy /auth/methods to API Gateway

standarius:


docker-compose.yml
```diff
@@ -73,6 +76,9 @@ services:
       - snjs
 
   api-gateway:
+    depends_on:
+      - syncing-server-ruby
+      - auth
     image: standardnotes/api-gateway:1.3.1
```
standarius, line 82: - [] this will need to be updated to a new version before merge

standarius: - [ ] also auth (line 95) will have to be updated

standarius: - [ ] and auth-worker (line 100)

docker-compose.yml
```diff
@@ -33,6 +33,10 @@ services:
       - snjs
 
   syncing-server-proxy:
+    depends_on:
+      - syncing-server-ruby
+      - syncing-server-js
+      - api-gateway
```
standarius, line 39: not sure if I got all the dependencies right, but with many tries I was unable to crash the setup on bootup -- whereas previously it did sometimes, if the dependencies booted in the wrong order

karolsojko: `depends_on` does not guarantee that the process in the container is started - it only secures the boot sequence of the containers.

Please use `wait-for.sh` for this purpose

karolsojko (approved):

docker-compose.yml
standarius: Addressed -- to make this work with nginx I had to create a Dockerfile for it.
Had trouble running it -- turns out there was a subtle bug in `wait-for.sh`, a consequence of the fact that in bash `exec $cmd` is not the same thing as `exec "$@"`, even if `cmd="$@"`.

karolsojko: hmm this change is not necessary I think ... as nginx will be out of the picture.

These 2 PRs are very tighly coupled as it turns out: https://github.com/standardnotes/snjs/pull/261

Not sure yet which one should go first

karolsojko:
@standarius can you rebase the branch so we can see the pr without conflicts?

karolsojko (approved):

PR #257, opened 2021-04-17
feat: ThemesActivated component action

johnny243:
Creates the new `ThemesActivated` component action. This action will be triggered after all themes have been activated by the component relay.


packages/snjs/lib/services/component_manager.ts
```diff
@@ -1268,6 +1270,13 @@ export class SNComponentManager extends PureService {
     this.desktopManager.installComponent(targetComponent);
   }
 
+  handleThemesActivatedMessage(
+    _sourceComponent: SNComponent,
+    _message: ComponentMessage
+  ) {
+    throw 'Must override SNComponentManager.handleThemesActivatedMessage';
```
mobitar, line 1277: I don't think web overrides ComponentManager. Only mobile does. So we have to make this a safe operation. Also, I don't think this function needs overriding. Should be a private function that simply sends the ComponentAction.ThemesActivated method to observers.

johnny243: In that case, we don't need to handle this action here and instead handle it within the `actionHandler` that is passed to `componentManager.registerHandler`:

- for web: https://github.com/standardnotes/web/blob/b93a35f9653b6d4b0d5bf37c1b207402b0b8aca2/app/assets/javascripts/views/editor/editor_view.ts#L1036
- for mobile: https://github.com/standardnotes/mobile/blob/c9c6db0ab2be2e23c2543b09ea7d828ed513d480/src/screens/Compose/Compose.tsx#L177

What do you think?

mobitar: Yup exactly 👍

mobitar (approved):

PR #261, opened 2021-04-20
feat: add new test suite setup without syncing server ruby

karolsojko:


karolsojko:
This removes the syncing-server-ruby completely out of the picture. 

`syncing-server-js` is fixed on `dev` version untill it's promoted to production - I will switch to a specific version number then.

baptiste-grob (approved):
Exciting stuff 👀

PR #265, opened 2021-04-22
feat: do not allow decrypted payloads from server

mobitar:


packages/snjs/lib/services/sync/account/response_resolver.ts
```diff
@@ -6,6 +7,31 @@ import { ImmutablePayloadCollection } from '@Protocol/collection/payload_collect
...
+function filterDisallowedPayloads(payloads: PurePayload[]): PurePayload[] {
+  return payloads.filter((payload) => {
+    const isDecrypted =
+      payload.format === PayloadFormat.DecryptedBareObject ||
+      payload.format === PayloadFormat.DecryptedBase64String;
+    const allowedToBeDecrypted = NonEncryptedTypes.includes(
+      payload.content_type
+    );
+
+    return !isDecrypted || (isDecrypted && allowedToBeDecrypted);
```
standarius, line 32: could be simplified to:

```js
    isEncrypted || NonEncryptedTypes.includes(
      payload.content_type
    )
```

where `isEncrypted` is:

```js
![PayloadFormat.DecryptedBareObject, PayloadFormat.DecryptedBase64String].includes(payload.format)
```

packages/snjs/lib/services/sync/account/response_resolver.ts
```diff
@@ -68,7 +94,7 @@ export class SyncResponseResolver {
 
     if (this.response.dataConflictPayloads.length > 0) {
       const collectionDataConflicts = await this.collectionByProcessingPayloads(
-        this.response.dataConflictPayloads,
+        filterDisallowedPayloads(this.response.dataConflictPayloads),
```
standarius, line 97: looks like the filtering shoud be moved to `collectionByProcessingPayloads`

packages/snjs/lib/services/sync/account/response_resolver.ts
mobitar: True, but I wonder what's easier to read for a human. I try to essentially model the var names to be like sentences.

packages/snjs/lib/services/sync/account/response_resolver.ts
```diff
@@ -93,7 +119,7 @@ export class SyncResponseResolver {
       this.baseCollection,
       collection,
       this.relatedCollectionSet,
-      this.historyMap,
```
standarius, line 96: ;(

mobitar: Wasn't me I swear!

packages/snjs/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@standardnotes/snjs",
-  "version": "2.0.76",
+  "version": "2.0.77",
```
standarius, line 3: semver nitpick (to ignore if this is a bugfix or we don't care): if this change is considered a new (backwards-compatible) feature (as `feat` in the commit suggests), the minor component of the version should be bumped rather than the patch component (which is for backwards-compatible bugfixes).

mobitar: Increasing anything but the patch version gives me anxiety 😥

packages/snjs/lib/services/sync/account/response_resolver.ts
standarius: another way to put this suggestion is:

```js
isEncrypted || isAllowedDecrypted
```

(where `isAllowedDecrypted` is the extracted includes check -- note however that with extracting there's no short-circuiting benefit -- but that may be irrelevant here)

So the sentences we're comparing are "not decrypted or (decrypted and allowed [to be] decrypted)" and "encrypted or allowed decrypted".

I think the latter is clearly easier to read.

packages/snjs/package.json
standarius: Looking at [the summary of semver](https://semver.org/#summary) to guide your decision may ease the anxiety. ;)

Anyway, adhering to semver is not strictly necessary for internal code. And in fuzzy cases it may indeed be difficult to make these decisions. I myself am not sure how to classify this change, so just pointing out, based on the commit message that minor might be more appropriate here.

In the end it's a matter of agreement between the library authors and users what the version means and what to expect when it changes.

test/sync_tests/online.test.js
```diff
@@ -1012,4 +1012,29 @@ describe('online syncing', function () {
     expect(this.application.apiService.session.accessToken).to.be.ok;
     expect(this.application.apiService.session.refreshToken).to.be.ok;
   });
+
+  it('should not allow receiving decrypted payloads from server', async function () {
+    const masterCollection = this.application.payloadManager.getMasterCollection();
+    const historyMap = this.application.historyManager.getHistoryMapCopy();
+    const note = await Factory.createSyncedNote(this.application);
+    this.expectedItemCount++;
```
standarius, line 1020: this looks dangerous
why do this?

packages/snjs/package.json
standarius: A good argument for being more aggressive in increasing the minor and major components: this way you err on the safe side -- the worst that happens is you communicate that you introduced a bigger or a breaking change when you didn't. This is better than the reverse: communicating that no breaking change was introduced when there was.

A node-specific argument for adhering to semver is that it is a common (and expected) practice -- `npm`s dependency, `node-semver` expects the `"version"` field from `package.json` to be at least formatted according to semver. Enforcing meaning is of course impractical.

So probably a safe default is to adhere to semver unless explicitly following a different policy.

packages/snjs/lib/services/sync/account/response_resolver.ts
```diff
@@ -6,6 +7,30 @@ import { ImmutablePayloadCollection } from '@Protocol/collection/payload_collect
...
+  ContentType.Mfa,
+  ContentType.ServerExtension,
+]);
+
+function filterDisallowedPayloads(payloads: PurePayload[]): PurePayload[] {
+  return payloads.filter((payload) => {
+    const isEncrypted = ![
+      PayloadFormat.DecryptedBareObject,
+      PayloadFormat.DecryptedBase64String,
+    ].includes(payload.format);
```
baptiste-grob, line 28: This has always been a weird brain twister for me 😅 I always end up translating it to _payload.format is neither DecryptedBareObject nor DecryptedBase64String_. But the code says it in a completely different order. I'm curious, how do you reason about code that you write in this way?

packages/snjs/lib/services/sync/account/response_resolver.ts
```diff
@@ -6,6 +7,30 @@ import { ImmutablePayloadCollection } from '@Protocol/collection/payload_collect
...
+    const isAllowedDecrypted = NonEncryptedTypes.includes(payload.content_type);
```
baptiste-grob, line 29: Interestingly though when the array has a name it's much more straightforward for me to understand. The grouping under a common name definitely helps.

packages/snjs/package.json
baptiste-grob: An argument I would make in favor of following semver: we have sometimes been stuck wanting to bundle an SNJS fix into a client, but there was a bigger feature in the commit tree as well that we didn't want to include. Had we used semver it would have been easier to branch off `master` and publish a minor version with only the fixes we needed.

test/sync_tests/online.test.js
baptiste-grob: If the sync server gets compromised and an attacker introduces a malicious payload to our items collection, this change would reject that payload as it couldn't have been encrypted with our local-only root key

standarius: Ah, you're answering about the entire pull request. This I understand.
What I'm asking about here is the line with `this.expectedItemCount++`. Seems like it's changing some state which doesn't seem to have anything to do with the rest of the test.
What is `expectedItemCount` used for?

packages/snjs/lib/services/sync/account/response_resolver.ts
standarius: Before applying the negation this can be read backwards as "format is one of x, y, z" -- with negation this becomes "format is not one of x, y, z".

What may be problematic is the negation that is applied last -- always gotta keep an eye on those.

test/sync_tests/online.test.js
mobitar: There are some tests in the afterEach block on the describe block. This I have already discovered to be poor form from the "sibling tests share state" thread so will be a pattern of the past, but for now, this is the way the tests in this file are described.

packages/snjs/lib/services/sync/account/response_resolver.ts
```diff
@@ -6,6 +7,30 @@ import { ImmutablePayloadCollection } from '@Protocol/collection/payload_collect
...
+ * These include server extensions (such as a note history endpoint), and
+ * multi-factor authentication items, which include a secret value that the server
+ * needs to be able to read in order to enforce.
+ */
+export const NonEncryptedTypes = Object.freeze([
+  ContentType.Mfa,
+  ContentType.ServerExtension,
+]);
+
+function filterDisallowedPayloads(payloads: PurePayload[]): PurePayload[] {
```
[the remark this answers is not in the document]
mobitar: For not being encrypted..

packages/snjs/package.json
mobitar: Agreed. But I would say this really isn't a major change nor a backward incompatible one. We don't rely on the functionality we're disabling. Just a preventative measure.

standarius: @mobitar sounds like increasing patch is fine in this case then -- it fixes a (potential) issue in a backwards-compatible manner (assuming nothing will break), and does not add any functionality

what may be considered confusing is the `feat` in the commit, which suggests a feature -- the intended meaning in the [Conventional Commits spec](https://www.conventionalcommits.org/en/v1.0.0/) is:

> feat: a commit of the type feat introduces a new feature to the codebase (this correlates with MINOR in Semantic Versioning).

now do we want to start using these in line with these intentions or do we agree that they only have provisional meaning for us (as seems to be the case so far)?

mobitar: Not sure, but will think harder about it next time.

PR #269, opened 2021-04-30
fix: check for undefined removeObserver

antsgar:


antsgar:
> `deinit()` being called twice on a `LiveItem` seems like a fault of the caller… Should this maybe throw/log an error instead? Can you provide some more context for this change?

Agreed, unfortunately I haven't got to the cause of why deinit is called again, so this definitely calls for a more in-depth fix. But I figured there'd be no harm still in checking for undefined before calling removeObserver to make it more fail-proof, seems like we are already doing this with other observers on SNJS.

mobitar:
The only harm would be that we've now suppressed the only warning that clients are doing something wrong 😅 Perhaps also console.error?

antsgar:
Added an error log 🙂 however I think maybe we can leave this PR open as draft until the issue has been investigated in more depth

packages/snjs/lib/models/live_item.ts
```diff
@@ -29,7 +29,11 @@ export class LiveItem<T extends SNItem> {
   }
 
   public deinit() {
-    this.removeObserver?.();
-    this.removeObserver = undefined;
+    if (!this.removeObserver) {
+      console.error('LiveItem removeObserver is undefined');
```
mobitar, line 33:
```suggestion
      console.error('A LiveItem is attempting to be deinited more than once.');
```

PR #273, opened 2021-05-03
feat: persist timestamps from server

mobitar:


karolsojko:
Rebased onto latest master with ss-js timestamp changes

PR #286, opened 2021-05-12
feat: adjust Auth SDK to the MFA implementation in Auth

standarius:
[ticket](https://app.asana.com/0/1199914526147392/1200322309691500/f)


packages/auth/src/Domain/User/AuthMethods.ts
```diff
@@ -1,7 +1,3 @@
-import { Uuid } from '../Uuid/Uuid'
-
 export type AuthMethods = {
-  totp?: {
-    mfaSecretUuid: Uuid,
-  },
+  totp?: true,
```
standarius, line 2: note: this got a little wonky (as it became a simple boolean), but is in line with the previous implementation.

this type ensures that either the `totp` prop does not exist at all in the object or is true if it does.
alternatively it could be a boolean and be required to always exist, but I didn't want to violate the policy of not exposing too much information from the endpoint that returns the auth methods. cc @mobitar

karolsojko (approved):

standarius:
Merging as this is blocking other tasks.

PR #287, opened 2021-05-12
feat: add methods for multiple selection

antsgar:
- Extracted some methods needed for multiple selection logic: getAuthorizedNotesForProtectedAction (not sure about the right naming) so that we only prompt for authentication challenge once and authorize a protected action on a batch of notes, and protectNotes and unprotectNotes to protect/unprotect in batch.
- Added SelectProtectedNote challenge reason and its corresponding text.

### DIFF
--- a/packages/snjs/lib/services/protection_service.ts
+++ b/packages/snjs/lib/services/protection_service.ts
@@ -171,7 +171,7 @@ export class SNProtectionService extends PureService<ProtectionEvent.SessionExpi
     }
   }
 
-  async getAuthorizedNotesForProtectedAction(
+  async authorizeProtectedActionForNotes(
     notes: SNNote[],
     challengeReason: ChallengeReason
   ): Promise<SNNote[]> {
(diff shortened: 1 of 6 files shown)


packages/snjs/lib/services/protection_service.ts
```diff
@@ -170,6 +171,46 @@ export class SNProtectionService extends PureService<ProtectionEvent.SessionExpi
     }
   }
 
+  async getAuthorizedNotesForProtectedAction(
```
mobitar, line 174: Hmm yeah I'm struggling with the naming of this function. So this function authorized an action to be performed on an array of notes, and the output is the array of notes for which the action was authorized? And the output length can be different from the input length?

In that case maybe just `authorizeNotesForProtectedAction` would work?

mobitar: Or maybe actually `authorizeProtectedActionForNotes`

mobitar (approved):

packages/snjs/lib/services/protection_service.ts
antsgar: Awesome! Yup that's the gist of it. I wasn't too fond of the name either, authorizeProtectedActionForNotes sounds good 🙂

PR #288, opened 2021-05-13
feat: add searchTags method

antsgar:
Add searchTags method which receives a search query and returns tags with titles matching that query, will be used for the autocomplete tags feature.


mobitar (approved):

packages/snjs/lib/services/item_manager.ts
```diff
@@ -721,6 +718,29 @@ export class ItemManager extends PureService {
...
+      .filter((tag) =>
+        tag.title
+          .toLowerCase()
+          .split(delimiter)
+          .some((component) => component.startsWith(searchQuery.toLowerCase()))
+      )
+      .sort(
+        typeof Intl !== 'undefined'
+          ? (a, b) =>
+              new Intl.Collator('en', { numeric: true }).compare(
```
mobitar, line 736: I read before I believe that Collator objects should be reused if possible?

baptiste-grob: Collator should also be safe to use at this point 🙂 https://caniuse.com/?search=collator

antsgar: It's safe for pretty much all browsers but Hermes doesn't have support for Intl yet, so we need the undefined check for mobile

packages/snjs/lib/services/item_manager.ts
```diff
@@ -721,6 +718,29 @@ export class ItemManager extends PureService {
...
+                a.title,
+                b.title
+              )
+          : (a, b) => (a.title > b.title ? 1 : -1)
+      );
+  }
```
antsgar, line 742: @mobitar would this be the change you're suggesting for reusing collator?

```suggestion
  public searchTags(searchQuery: string): SNTag[] {
    const delimiter = '.';
    const collator = new Intl.Collator('en', { numeric: true });
    return this.tags
      .filter((tag) =>
        tag.title
          .toLowerCase()
          .split(delimiter)
          .some((component) => component.startsWith(searchQuery.toLowerCase()))
      )
      .sort(
        typeof Intl !== 'undefined'
          ? (a, b) =>
              collator.compare(
                a.title,
                b.title
              )
          : (a, b) => (a.title > b.title ? 1 : -1)
      );
  }
```

baptiste-grob: Since it's a stateless object you could potentially move it even higher up, as a constant

baptiste-grob: You likely could use `a.title.localeCompare(b.title)` instead of `a.title > b.title` 🙂

packages/snjs/lib/services/item_manager.ts
```diff
@@ -742,7 +742,7 @@ export class ItemManager extends PureService {
                 a.title,
                 b.title
               )
-          : (a, b) => a.title.localeCompare(b.title)
+          : (a, b) => a.title.localeCompare(b.title, 'en', { numeric: true })
```
mobitar, line 745: Nice. We wanted to add this as well for sorting by title in the notes list. Can be done in a separate PR though.

baptiste-grob (approved):

PR #304, opened 2021-05-24
feat: use regex and add note parameter for searchTags

antsgar:
- Use regex for query matching, so that a result will be returned if either the title starts with the query, or the query is present after a delimiter. This means if we have a tag with a "parent.child.grandchild" title, searching for "parent.child" and "child.grandchild" will both return it as a result.
- Add optional note parameter to searchTags method, so that when this parameter is passed, tags associated with that note will not be returned in results. This will allow us to show only tags not currently added to the note in the search dropdown.

### DIFF
--- a/packages/snjs/lib/application.ts
+++ b/packages/snjs/lib/application.ts
@@ -773,6 +773,12 @@ export class SNApplication {
     return this.itemManager.findTagByTitle(title);
   }
 
+  /**
+   * Finds tags with title or component starting with a search query and (optionally) not associated with a note
+   * @param searchQuery - The query string to match
+   * @param note - The note whose tags should be omitted from results
+   * @returns Array containing tags matching search query and not associated with note
+   */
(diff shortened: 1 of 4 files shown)


packages/snjs/lib/application.ts
```diff
@@ -773,8 +773,8 @@ export class SNApplication {
     return this.itemManager.findTagByTitle(title);
   }
 
-  public searchTags(searchQuery: string): SNTag[] {
-    return this.itemManager.searchTags(searchQuery);
+  public searchTags(searchQuery: string, note?: SNNote): SNTag[] {
```
mobitar, line 776: Hmm it might not be clear to a caller how nite is being used? I think I can take a good guess but maybe an @param comment doc would be helpful.

mobitar (approved):

packages/snjs/lib/application.ts
antsgar: Good call! I've improved the comment doc and added it to the application method too

PR #305, opened 2021-05-25
fix: use created_at for unsynced session history entries

baptiste-grob:
See https://github.com/standardnotes/mobile/issues/432

### DIFF
--- a/packages/snjs/lib/services/history/entries/history_entry.ts
+++ b/packages/snjs/lib/services/history/entries/history_entry.ts
@@ -12,10 +12,7 @@ export class HistoryEntry {
   protected readonly hasPreviousEntry: boolean;
 
   constructor(payload: SurePayload, previousEntry?: HistoryEntry) {
-    const updated_at = payload.serverUpdatedAt ?? new Date();
-    this.payload = CopyPayload(payload, {
-      updated_at: updated_at,
-    }) as SurePayload;
+    this.payload = CopyPayload(payload) as SurePayload;
     this.previousEntry = previousEntry;
     this.hasPreviousEntry = !isNullOrUndefined(previousEntry);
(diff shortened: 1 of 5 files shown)


packages/snjs/lib/services/history/entries/history_entry.ts
```diff
@@ -12,10 +12,7 @@ export class HistoryEntry {
   protected readonly hasPreviousEntry: boolean;
 
   constructor(payload: SurePayload, previousEntry?: HistoryEntry) {
-    const updated_at = payload.serverUpdatedAt ?? new Date();
```
mobitar, line 15: Hmm, here we are creating new Date(), which is the present date, but in the constructor of pure_payload we are defaulting to new Date(0), which is 1970. Is this intended? So seems like here after this change if updated_at is nil then entry will default to 1970 rather than present.

baptiste-grob: There's a bug with this line in that `payload.serverUpdatedAt` is always truthy, because the pure_payload constructor always creates a date object. So `updated_at` can never be `new Date()`.

I don't know what the original intent was, but I think a history entry making up dates is more confusing than helpful. `updated_at` being equal to the time of construction does not make sense with how the rest of the code treats that field

Talking more broadly I'm not sure setting a payload's `updated_at` to 1/1/1970 by default is a very accurate way to describe an unsynced payload, but I may be missing something

mobitar: I think the intent is that if creating a new entry, the present time is likely the right time for its creation. But this applies only to the offline use case. I suppose as long as you've tested this functionality in the web app while not signed into an account, and it shows the right date, then all is good.

mobitar (approved):

PR #306, opened 2021-05-25
feat: natural sort tags and notes

antsgar:
- Added naturalSort util which uses Intl.Collator (if available), and if not, locale.compare.
- Modified searchTags to use naturalSort util, to use the case-insensitive flag in the regex instead of transforming strings using toLowerCase(), and to compare tag UUID instead of the entire tag when filtering out tags already in a note.
- Added getSortedTagsForNote method, which returns tags for a note in natural order. This will be used to display tags in the current note, now that we will no longer show them joined in a string.
- Added { numeric: true } parameter to item sorting by title, so notes ordered by title are sorted in natural order

### DIFF
--- a/packages/snjs/lib/utils.ts
+++ b/packages/snjs/lib/utils.ts
@@ -553,19 +553,25 @@ export function isSameDay(dateA: Date, dateB: Date) {
    * @param direction - The sorting direction, either ascending (default) or descending
    * @returns Array of objects sorted in natural order
    */
-export function naturalSort(items: any[], property: string, direction: 'asc' | 'desc' = 'asc'): any[] {
+export function naturalSort<T extends AnyRecord>(
+  items: T[],
+  property: keyof T,
+  direction: 'asc' | 'desc' = 'asc'
+): T[] {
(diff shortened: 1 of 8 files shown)


packages/snjs/lib/utils.ts
```diff
@@ -540,3 +545,27 @@ export function isSameDay(dateA: Date, dateB: Date) {
...
 }
+
+ /**
+   * Sorts an array of objects in natural order
+   * @param items - The array of objects to sort
+   * @param property - The objects' property to sort by
+   * @param direction - The sorting direction, either ascending (default) or descending
+   * @returns Array of objects sorted in natural order
+   */
+export function naturalSort(items: any[], property: string, direction: 'asc' | 'desc' = 'asc'): any[] {
```
mobitar, line 556: Can use generics here? <T> and keyof T for property.

mobitar (approved):

packages/snjs/lib/utils.ts
antsgar: Done! 👌

PR #318, opened 2021-06-03
feat: add getParentTags method

antsgar:
Added getParentTags method which returns all parents for a specific tag. This will be used for adding all parents on top of a tag when clicking on it on a note.

### DIFF
--- a/packages/snjs/lib/services/item_manager.ts
+++ b/packages/snjs/lib/services/item_manager.ts
@@ -743,6 +743,23 @@ export class ItemManager extends PureService {
   }
 
   /**
+   * Returns all parents for a tag
+   * @param tag - The tag for which parents need to be found
+   * @returns Array containing all parent tags
+   */
+  public getParentTags(tag: SNTag): SNTag[] {
+    const delimiter = '.';
+    const tagComponents = tag.title.split(delimiter);
+    const childTagTitle = tagComponents[tagComponents.length - 1];
+    return this.tags.filter((t) => {
+      const regex = new RegExp(
+        `${delimiter}${childTagTitle}$`
+      );
+      return regex.test(t.title);
+    });
+  }
+
+  /**
    * Get tags for a note sorted in natural order
    * @param note - The note whose tags will be returned
    * @returns Array containing tags associated with a note
--- a/test/item_manager.test.js
+++ b/test/item_manager.test.js
@@ -582,5 +582,21 @@ describe('item manager', function () {
       expect(results[2].title).to.equal(tags[3].title);
       expect(results[3].title).to.equal(tags[2].title);
     })
+  });
+
+  describe('getParentTags', async function () {
+    it('should return parent tags for a tag', async function () {
+      const parentTags = [
+        await this.itemManager.findOrCreateTagByTitle('parent.child.grandchild'),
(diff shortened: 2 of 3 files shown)


test/item_manager.test.js
```diff
@@ -582,5 +582,21 @@ describe('item manager', function () {
       expect(results[2].title).to.equal(tags[3].title);
       expect(results[3].title).to.equal(tags[2].title);
     })
+  });
+
+  describe('getParentTags', async function () {
+    it('should return parent tags for a tag', async function () {
+      const parentTags = [
+        await this.itemManager.findOrCreateTagByTitle('parent.child.grandchild'),
```
mobitar, line 590: Hmm..I don't think this is right. Parent tags of `parent.child.grandchild` would be named `parent.child` and `parent`. So parent tags don't include `grandchild`.

packages/snjs/lib/services/item_manager.ts
```diff
@@ -742,6 +742,23 @@ export class ItemManager extends PureService {
     );
   }
 
+  /**
+   * Returns all parents for a tag
+   * @param tag - The tag for which parents need to be found
+   * @returns Array containing all parent tags
+   */
+  public getParentTags(tag: SNTag): SNTag[] {
```
mobitar, line 750: As-is I don't think this function works as expected.

For `foo.bar.tar.zar`, you'd want to return the following parents:

- foo.bar.tar
- foo.bar
- foo

So perhaps you'd return all tags that startWith `foo.`? On the other hand this would return all unrelated children as well, like `foo.bar.tar.sar.zar`. So instead you'd have to just scan the input string itself and construct all possible parent title combinations.

Then you'd name this function something like `getParentChain`, since a tag can only have 1 parent.

test/item_manager.test.js
antsgar: Oops you're right, this doesn't make sense at all. Will fix it now!

packages/snjs/lib/services/item_manager.ts
antsgar: I've updated the logic, please let me know if it makes sense now, not sure what my brain was thinking the first time around 😆

antsgar:
> What if you first recursively built the various titles, then at the end you only loop through all tags once.
> 
> Because right now you are looping through all tags for every level, which isn't the most performant route, especially if a user say has 1000 tags (you'd be surprised 😄)

Done! 😄

packages/snjs/lib/services/item_manager.ts
```diff
@@ -734,14 +734,45 @@ export class ItemManager extends PureService {
...
+    const parentTagsTitles: string[] = [];
+
+    const getImmediateParent = () => {
+      if (tagComponents.length > 1) {
+        // Remove last component
+        tagComponents.splice(-1, 1);
+        // Get immediate parent title and add it to chain
+        const immediateParentTitle = tagComponents.join(delimiter);
+        parentTagsTitles.push(immediateParentTitle);
+        // Get parent of this parent recursively
```
mobitar, line 764: I think the code is actually very readable and easy to follow so these comments aren't necessary :)

mobitar (approved):

PR #319, opened 2021-06-07
feat: add alternative host and switch sign out to API v1

karolsojko:
- added the `alternativeHost` property so that we can switch endpoints one by one to an upgraded version of the API. Will come in handy when switching to for example `v2` as well in future. The naming could be better than `alternativeHost` I guess so I'm open for propositions.
- switched the sign out action to utilize API `v1` instead of `v0`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -54,7 +54,7 @@ services:
       - snjs
 
   api-gateway:
-    image: standardnotes/api-gateway:1.14.0
+    image: standardnotes/api-gateway:1.15.4
     depends_on:
       - syncing-server-js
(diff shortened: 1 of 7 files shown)


karolsojko:
A fix on API Gateway side, allowing expired access tokens to sign out was required: https://github.com/standardnotes/api-gateway/commit/bb345a3a3aedd015d1fd95c176035ad98bb2424e

mobitar (approved):
Perhaps `versionNextHost`?

karolsojko:
Agreed with Mo on `nextVersionHost` in DM

PR #322, opened 2021-06-09
feat: make tags search case sensitive

antsgar:
Since tag titles are case sensitive (there can be two different tags with only a case difference), search should also be.


antsgar:
@mobitar that's interesting, not sure if there might be a use case for it for some people?

mobitar:
I doubt it. Sounds like the check we do before creating a tag just forgot to be case insensitive.

antsgar:
Closing and opening a different PR to make tag titles case insensitive.

PR #323, opened 2021-06-09
feat: make tag titles case insensitive

antsgar:


packages/snjs/lib/services/item_manager.ts
```diff
@@ -714,8 +714,8 @@ export class ItemManager extends PureService {
   /**
    * Finds the first tag matching a given title
    */
-  public findTagByTitle(title: string) {
-    return searchArray(this.tags, { title: title });
+  public findTagByTitle(title: string): SNTag | undefined {
+    return this.tags.find((tag) => tag.title.toLowerCase() === title.toLowerCase());
```
mobitar, line 718: I would toLowerCase `title` param outside of the loop to avoid doing the same computation many times inside the loop.

mobitar (approved):

packages/snjs/lib/services/item_manager.ts
antsgar: Good call, done!

PR #325, opened 2021-06-10
feat: migrate getItemRevisions to api/v1

antsgar:


karolsojko (approved):
@antsgar can you bump the version of the package? As a rule of thumb I think `2.6.3` would be good

antsgar:
@karolsojko yup I was waiting for the other PRs to be merged to figure out the right version number 🙂

PR #346, opened 2021-07-01
fix: set minimum passcode length

antsgar:


mobitar (approved):
Test would be ideal if possible. "should fail when setting 0 character passcode" etc

antsgar:
> Test would be ideal if possible. "should fail when setting 0 character passcode" etc

Done! 🙂

PR #347, opened 2021-07-01
fix: add check for undefined passcode

antsgar:


packages/snjs/lib/services/credential_service.ts
```diff
@@ -404,7 +404,7 @@ export class SNCredentialService extends PureService<AccountEvent> {
   }
 
   public async addPasscode(passcode: string): Promise<boolean> {
-    if (passcode.length < MINIMUM_PASSCODE_LENGTH) {
+    if (!passcode || passcode.length < MINIMUM_PASSCODE_LENGTH) {
```
mobitar, line 407: Hmm are we not trusting TypeScript compiler enough here?

antsgar: Right now when the inputs were empty on web, we were passing the passcode parameter as undefined, not an empty string. But you're 100% right, I've re-checked the web code and the problem is we need better type handling on the client side.

antsgar: Closing the PR, thanks for the feedback! 🙂

PR #355, opened 2021-07-08
feat: change permission and role types

antsgar:


packages/auth/src/Domain/Role/Role.ts
```diff
@@ -1,6 +1,13 @@
-export enum Role {
+import { Uuid } from '../Uuid/Uuid'
+
+export enum RoleName {
   User = 'USER',
```
karolsojko, line 4: can we change this one to `BASIC_USER`?

packages/auth/src/Domain/Permission/Permission.ts
```diff
@@ -1,5 +1,12 @@
-export enum Permission {
+import { Uuid } from '../Uuid/Uuid'
+
+export enum PermissionName {
   SyncItems = 'SYNC_ITEMS',
```
karolsojko, line 4: could we change the permission names to the ones that are used? 
https://github.com/standardnotes/auth/blob/develop/migrations/1624434102642-update_roles_and_permissions.ts#L17-L46

karolsojko (changes requested):

packages/auth/src/Domain/Permission/Permission.ts
antsgar: Done!

packages/auth/src/Domain/Role/Role.ts
antsgar: Done!

karolsojko (approved):

PR #360, opened 2021-07-16
feat: add subscription events

karolsojko:


mobitar (approved):

packages/domain-events/src/Domain/Event/SubscriptionRenewedEvent.ts
```diff
@@ -0,0 +1,8 @@
+import { DomainEventInterface } from './DomainEventInterface'
+
+import { SubscriptionCancelledEventPayload } from './SubscriptionCancelledEventPayload'
+
+export interface SubscriptionCancelledEvent extends DomainEventInterface {
```
antsgar, line 5: @karolsojko just noticed SubscriptionCancelledEvent and SubscriptionRenewedEvent are each on the other's corresponding file

karolsojko: ah my bad - copy pasting :) will fix

PR #362, opened 2021-07-19
feat: add feature type

antsgar:


mobitar (approved):

packages/auth/src/Domain/Feature/Feature.ts
```diff
@@ -0,0 +1,67 @@
+import { ComponentArea, ContentType } from '@standardnotes/snjs';
+
+export enum FeatureName {
```
karolsojko, line 3: Hmm maybe this should be `PermissionName` if they are the same ?

packages/auth/src/Domain/Feature/Feature.ts
```diff
@@ -0,0 +1,67 @@
...
+  identifier: string;
+  content_type: ContentType;
+  area: ComponentArea;
+  layerable: boolean;
+  no_mobile: boolean;
+  version: string;
+  description: string;
+  url: string;
+  download_url: string;
+  marketing_url: string;
```
karolsojko, line 57: can those be camelCase or is there a reason for snake here?

packages/auth/package.json
```diff
@@ -27,6 +27,7 @@
   "dependencies": {},
   "devDependencies": {
     "@standardnotes/config": "^2.0.0",
+    "@standardnotes/snjs": "2.7.20",
```
karolsojko, line 30: hmm wondering if that won't be circular - snjs will depend on auth and auth will depend on snjs. 

Also seems an overkill for auth to require the whole SNJS. 

We have to figure out this dependecies mashup I think - wdyt?

karolsojko (changes requested):

packages/auth/src/Domain/Feature/Feature.ts
antsgar: Sure thing! I was just basing myself on the current shape from the extensions server but this can be camelCase no problem 🙂

packages/auth/src/Domain/Feature/Feature.ts
antsgar: I'm thinking we'll get rid of the Permission and Role types on SNJS once clients no longer deal with permissions?

packages/auth/src/Domain/Feature/Feature.ts
```diff
@@ -0,0 +1,67 @@
...
+  Circle = 'circle',
+  Svg = 'svg'
+}
+
+enum Flag {
+  New = 'New',
+  Deprecated = 'Deprecated'
+}
+
+export type Feature = {
```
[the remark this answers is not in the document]
antsgar: Will do!

packages/auth/package.json
antsgar: Yup true story! Maybe we could have a subset of ContentType and ComponentArea specified on auth? Con for that would be we'd have to make changes on both packages when we add new options

packages/auth/src/Domain/Feature/Feature.ts
karolsojko: yeah, the server uses the `auth` package

packages/auth/package.json
karolsojko: hmm `auth` doesn't seem to be a good place for `ContentType` and `ComponentArea`. 

I'd suggest extract those into a new package, and make both `auth` and `snjs` dependant on that new package.

Lacking a good name for the package - @mobitar any ideas :) ?

karolsojko: maybe `content` or `item` ? Trying to come up with something more general that could umbrella at least a couple of things

antsgar: Perhaps we could move all these types to a separate features package?

karolsojko: good idea! do you need help in creating the package :) ?

antsgar: Created the new package, left ContentType and ComponentArea and made the new features package dependent on both auth and SNJS (but neither of them depends on features). Let me know if it looks alright 🙂

packages/snjs/package.json
```diff
@@ -21,7 +21,7 @@
     "lint": "yarn lint:tsc && yarn lint:eslint",
     "lint:eslint": "eslint --ext .ts lib/",
     "lint:tsc": "tsc --noEmit --emitDeclarationOnly false --project lib/tsconfig.json",
-    "test": "jest"
+    "test": "jest"
```
karolsojko, line 24: not needed whitespace

packages/features/src/Domain/Feature/Feature.ts
```diff
@@ -0,0 +1,28 @@
+import { PermissionName } from '@standardnotes/auth'
+import { ComponentArea } from '@standardnotes/snjs'
+import { ContentType } from '@standardnotes/snjs'
+import { DockIconType } from './DockIconType'
+import { Flag } from './Flag'
+
+export type Feature = {
+  name: PermissionName;
+  identifier: string;
+  contentType: ContentType
```
karolsojko, line 10: hmm maybe it make sense to move `ContentType` and `ComponentArea` to `features` package? or extract it to a `content` or `items` package?

The reason for this is that `features` we are not avoiding the circular dependency now:

- when you want to install features, it will install the whole SNJS
- `snjs` will require `features` at some point - and `features` already requires `snjs`
- the server applications will require `features` but not the entire snjs.

wdyt?

karolsojko (changes requested):

karolsojko:
One more requirement you also need to add the package to Dockerfile: https://github.com/standardnotes/snjs/blob/master/Dockerfile#L21-L25

packages/features/src/Domain/Feature/Feature.ts
antsgar: Ah that makes sense, I tried that option at first but wasn't sure because the whole component model lives on SNJS, so it felt a bit weird to have just ComponentArea live outside of it. Also unit tests don't pass when importing something from an external dependency on a model, but I'm sure we should be able to fix that

antsgar: Done! 🙂 ComponentArea and ContentType now live in features. snjs depends on features and auth, features depends on auth and auth depends on none

karolsojko (approved):
Just a note.

Not sure if this PR should be split into 2:
- first publish the features package
- then make SNJS dependant on features and remove ComponentArea and ContentType there

The reason is that SNJS will try to install a package that does not exist yet. Just a hunch so not entirely sure that this is the case - let's try with just this one PR - if it fails you'll know what to do in order to fix ;)

antsgar:
@karolsojko good point! Will give it a shot and if it doesn't work split the changes 🙂

PR #363, opened 2021-07-20
feat: add user role changed event for websockets purposes

karolsojko:
I was thinking about the events structure that we are going to utilize in the websockets communication and I think the events that we already use for SNS/SQS are good for using in websockets as well. So it's enough if we publish a `USER_ROLE_CHANGED` event via websockets and the client will know to re-fresh features based on that. @antsgar wdyt?


antsgar (approved):
@karolsojko sounds good! 🙂

PR #366, opened 2021-07-29
feat: remove fromRole from user role changed event payload

antsgar:


packages/domain-events/src/Domain/Event/UserRoleChangedEventPayload.ts
```diff
@@ -3,7 +3,6 @@ import { RoleName } from '@standardnotes/auth'
 export interface UserRoleChangedEventPayload {
   userUuid: string
   email: string
-  fromRole: RoleName
   toRole: RoleName
```
karolsojko, line 6: can we change this one to just `role` ?

karolsojko (approved):

karolsojko (approved):

PR #368, opened 2021-08-02
feat: add fromSubscriptionName to purchase and refund events

antsgar:


packages/domain-events/src/Domain/Event/SubscriptionPurchasedEventPayload.ts
```diff
@@ -2,7 +2,8 @@ import { SubscriptionName } from '@standardnotes/auth'
 
 export interface SubscriptionPurchasedEventPayload {
   userEmail: string
-  subscriptionName: SubscriptionName
+  fromSubscriptionName: SubscriptionName
```
karolsojko, line 5: from can be undefined right ? - purchasing a subscription from not having one
```suggestion
  fromSubscriptionName?: SubscriptionName
```

packages/domain-events/src/Domain/Event/SubscriptionRefundedEventPayload.ts
```diff
@@ -2,6 +2,7 @@ import { SubscriptionName } from '@standardnotes/auth'
 
 export interface SubscriptionRefundedEventPayload {
   userEmail: string
-  subscriptionName: SubscriptionName
+  fromSubscriptionName: SubscriptionName
+  toSubscriptionName: SubscriptionName
```
karolsojko, line 6: hmm if refunding we are cancelling a subscription so to say ... so no 'toSubscriptio` is needed right?

packages/domain-events/src/Domain/Event/SubscriptionPurchasedEventPayload.ts
antsgar: The fromSubscriptionName in that case should be CorePlan, correct? That way we would know we need to remove the CoreUser role from the user

packages/domain-events/src/Domain/Event/SubscriptionRefundedEventPayload.ts
antsgar: Same as above, the toSubscriptionName should be CorePlan. Does that make sense?

packages/domain-events/src/Domain/Event/SubscriptionPurchasedEventPayload.ts
karolsojko: hmm not really ... a basic user: https://github.com/standardnotes/snjs/blob/master/packages/auth/src/Domain/Role/RoleName.ts#L2

can buy a core plan: https://github.com/standardnotes/snjs/blob/master/packages/auth/src/Domain/Subscription/SubscriptionName.ts#L2

so it means that his subscription was null until now.

packages/domain-events/src/Domain/Event/SubscriptionRefundedEventPayload.ts
karolsojko: When someone cancels a core subscription they can get back to a basic user with no subscription

packages/domain-events/src/Domain/Event/SubscriptionPurchasedEventPayload.ts
antsgar: Ah thanks, that clears everything out! For some reason I was confusing Core with Basic thinking they were the same thing. Will update this

packages/domain-events/src/Domain/Event/SubscriptionRefundedEventPayload.ts
antsgar: @karolsojko come to think of it, it's the same case as above but reversed, so should be optional I think? A user can have for example a Plus subscription, buy a Pro subscription and then refund it. In that case, the user would go from Pro to Plus. Is that correct?

antsgar: @karolsojko actually I just realized we may not need this at all, how about we just add the new user role when a subscription is purchased and remove it if it's refunded? That way if a user doesn't have a subscription and a Pro plan is purchased for example, the user would have both the BasicUser and ProUser roles, but I think that'd be okay. Wdyt?

antsgar: Made the described change on the auth PR so will close this for now, will re-open if necessary 🙂

PR #369, opened 2021-08-03
feat: add features data

vardan-arm:
- make some feature fields optional
  - `expiresAt` is also optional, since it isn't required when getting features list for static website, for example
- some field values are placeholders, the missing information will be added later


karolsojko (approved):

karolsojko:
please bump the version in package.json in order for this to get published after merge

vardan-arm:
> please bump the version in package.json in order for this to get published after merge

Sure, I assume this should be minor bump right, so the new version will be `1.1.0`?

PR #371, opened 2021-08-04
fix: make features package independent of auth

karolsojko:
as the only thing that `features` package is depending on in `auth` is the permissions - which is a descriptor of a feature, it makes more sense to move the permissions model to the features package to make it independent of auth and avoid potential future circular dependencies


yarn.lock
```diff
@@ -2491,6 +2491,11 @@
     "@typescript-eslint/eslint-plugin" "^4.14.0"
     "@typescript-eslint/parser" "^4.14.0"
 
+"@standardnotes/features@1.0.0":
```
gorjan5sk, line 2494: I'm guessing we'll update the version when the dependents are updated?

karolsojko: this is for `@standardnotes/snjs` package ... no need to update the features dependcy version in there yet, but we can do it soon

vardan-arm:
Shouldn't we also remove Permissions-related files from `packages/auth/src`?

karolsojko:
@vardan-arm yes, coming in next PR :)

PR #372, opened 2021-08-04
feat: extract settings and common package

karolsojko:


packages/settings/src/Domain/Setting/SettingName.ts
```diff
@@ -1,3 +1,3 @@
-export enum MfaSetting {
+export enum SettingName {
   MfaSecret = 'MFA_SECRET',
 }
```
gorjan5sk, line 3:
```suggestion
/**
 * Enumerate all Settings with their names as identifiers
 *
 * Extend by adding more enum members
 */
export enum Setting {
  MfaSecret = 'MFA_SECRET',
}

/**
 * Describe Setting payloads (contents) matched with Setting enum keys
 *
 * Extend by adding more properties with keys from Setting enum and value with the corresponding Payload
 */
export type SettingPayload = {
  [Setting.MfaSecret]: { secret: string };
};

```

gorjan5sk: You beat me to the refactor. What do you think about this ^?

packages/auth/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@standardnotes/auth",
-  "version": "3.2.0",
```
gorjan5sk, line 3: This would be a breaking change for the dependent `auth` repo. I think we should change the MINOR version at least?

vardan-arm (approved):

packages/settings/src/Domain/Setting/SettingName.ts
karolsojko: hmm not sure if we need to define the payload like that - but feel free to add it another PR.

Additionally keep in mind to keep 1 type/class/interface per file

packages/auth/package.json
karolsojko: good point

gorjan5sk (approved):

PR #373, opened 2021-08-04
fix: replace permissions with features in auth token

karolsojko:


packages/auth/src/Domain/Token/Token.ts
```diff
@@ -12,12 +15,6 @@ export type Token = {
...
-  roles: Array<{
-    uuid: Uuid
-    name: string
-  }>,
-  permissions: Array<{
-    uuid: Uuid
-    name: string
-  }>,
+  roles: Array<Role>,
+  features: Array<Feature>,
```
vardan-arm, line 19: Just to double check - is it safe to replace `permissions` by `features` here?

karolsojko: yes. This does not take effect until `auth` and `api-gateway` update their version of the `@standardnotes/auth` package

PR #374, opened 2021-08-04
feat: implement SettingsService

gorjan5sk:


packages/snjs/lib/services/settings_service/SNSettingsService.ts
```diff
@@ -0,0 +1,69 @@
...
+  private get user(): User {
+    return this.user!;
+  }
+
+  private get provider(): SettingsProvider {
+    return this._provider!;
+  }
+
+  public async initializeFromDisk() {
+    this._user = await this.storageService!.getValue(StorageKey.User);
```
mobitar, line 32: Don't think we should be duplicating this from session_manager. Should request it from that service instead.

packages/snjs/lib/services/settings_service/SettingsCache.ts
```diff
@@ -0,0 +1,78 @@
...
+    this.lastUpdateMap.set(name, this.currentTime());
+  }
+
+  private async setAll(payload: Partial<SettingPayload>) {
+    this.cache = payload;
+    this.cache = await this.provider.listSettings();
+
+    const now = this.currentTime();
+    this.lastUpdate = now;
+    for (let k of this.lastUpdateMap.keys()) {
```
mobitar, line 29:
```suggestion
    for (const key of this.lastUpdateMap.keys()) {
```

We try to avoid single letter vars to make code more readable and searchable.

packages/snjs/lib/services/settings_service/SNSettingsService.ts
gorjan5sk: Fixed.

gorjan5sk:
Closing in favor upcoming smaller PRs with integration tests.

PR #377, opened 2021-08-06
feat: add settings payload domain definition

gorjan5sk:


packages/settings/src/Domain/Setting/SettingName.ts
```diff
@@ -1,3 +1,9 @@
+/**
+ * Enumerate all Settings with their names as identifiers
```
karolsojko, line 2: Not necessary comment. Does not add value.

packages/settings/src/Domain/Setting/SettingPayload.ts
```diff
@@ -0,0 +1,11 @@
+import { SettingName } from './SettingName'
+
+/**
+ * Describe Setting payloads (contents) matched with Setting enum keys
+ *
```
karolsojko, line 5: Same

karolsojko (approved):
Could you just remove the comments? They don't add any value in here

gorjan5sk:
@karolsojko removed comments. Merging! :)

PR #380, opened 2021-08-10
feat: update user role changed event

antsgar:
Update user role changed event to include all user roles instead of only the one that changed (otherwise we have no way of knowing if the role has been added or removed)


packages/domain-events/src/Domain/Event/UserRolesChangedEvent.ts
```diff
@@ -0,0 +1,7 @@
+import { DomainEventInterface } from './DomainEventInterface'
+import { UserRolesChangedEventPayload } from './UserRolesChangedEventPayload'
+
+export interface UserRolesChangedEvent extends DomainEventInterface {
+  type: 'USER_ROLE_CHANGED'
```
karolsojko, line 5:
```suggestion
  type: 'USER_ROLES_CHANGED'
```

packages/domain-events/src/Domain/Event/UserRolesChangedEventPayload.ts
```diff
@@ -0,0 +1,8 @@
+import { Role } from '@standardnotes/auth'
+
+export interface UserRolesChangedEventPayload {
+  userUuid: string
+  email: string
+  roles: Role[]
```
karolsojko, line 6:
```suggestion
  currentRoles: Role[]
```

karolsojko (changes requested):

antsgar:
Closing this one in favor of #381

PR #383, opened 2021-08-11
feat: revert to returning just user roles on event and auth token

antsgar:


packages/auth/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@standardnotes/auth",
-  "version": "3.6.1",
+  "version": "3.7.0",
   "engines": {
```
vardan-arm, line 4: Maybe it makes sense to remove `@standardnotes/features` [dependency](https://github.com/standardnotes/snjs/blob/master/packages/auth/package.json#L29) at all, since we won't be using it with these changes anymore?

karolsojko: hmm good point

antsgar: Thanks @vardan-arm! 🙂

PR #384, opened 2021-08-12
feat: settings service 

gorjan5sk:
- Implement SNSettingsService for reading, updating, deleting and listing settings
- Implement test for SNSettingsService

### DIFF
--- a/packages/snjs/lib/application.ts
+++ b/packages/snjs/lib/application.ts
@@ -101,6 +101,8 @@ import { SNPermissionsService } from './services/permissions_service';
 import { ProtectionEvent } from './services/protection_service';
 import { PermissionName } from '@standardnotes/auth';
 import { RemoteSession } from '.';
+import { SettingName } from '@standardnotes/settings';
+import { SNSettingsService } from './services/settings_service';
 
 /** How often to automatically sync, in milliseconds */
 const DEFAULT_AUTO_SYNC_INTERVAL = 30_000;
@@ -142,6 +144,7 @@ export class SNApplication {
   private preferencesService!: SNPreferencesService;
   private permissionsService!: SNPermissionsService;
   private credentialService!: SNCredentialService;
+  private settingsService!: SNSettingsService;
 
   private eventHandlers: ApplicationObserver[] = [];
   private services: PureService<any, any>[] = [];
@@ -187,7 +190,7 @@ export class SNApplication {
     public identifier: ApplicationIdentifier,
     private swapClasses: { swap: any; with: any }[],
     private defaultHost: string,
-    private webSocketUrl?: string,
+    private webSocketUrl?: string
   ) {
     if (!SNLog.onLog) {
       throw Error('SNLog.onLog must be set.');
@@ -290,6 +293,7 @@ export class SNApplication {
     await this.permissionsService.loadWebSocketUrl();
     await this.sessionManager.initializeFromDisk();
     this.historyManager.initializeFromDisk();
+    this.settingsService.initializeFromDisk();
 
     this.launched = true;
     await this.notifyEvent(ApplicationEvent.Launched);
@@ -517,7 +521,9 @@ export class SNApplication {
     return this.syncService.getStatus();
   }
 
-public getSessions(): Promise<(HttpResponse & { data: RemoteSession[] }) | HttpResponse> {
+  public getSessions(): Promise<
+    (HttpResponse & { data: RemoteSession[] }) | HttpResponse
+  > {
     return this.sessionManager.getSessionsList();
   }
 
@@ -800,7 +806,7 @@ public getSessions(): Promise<(HttpResponse & { data: RemoteSession[] }) | HttpR
    * @param tag - The tag for which descendants need to be found
    * @returns Array containing all descendant tags
    */
-   public getTagDescendants(tag: SNTag): SNTag[] {
+  public getTagDescendants(tag: SNTag): SNTag[] {
     return this.itemManager.getTagDescendants(tag);
   }
 
@@ -1441,6 +1447,22 @@ public getSessions(): Promise<(HttpResponse & { data: RemoteSession[] }) | HttpR
     }
   }
 
+  public async listSettings() {
+    return this.settingsService.settings().listSettings();
+  }
+
+  public async getSetting(name: SettingName) {
+    return this.settingsService.settings().getSetting(name);
+  }
+
+  public async updateSetting(name: SettingName, payload: string) {
+    return this.settingsService.settings().updateSetting(name, payload);
+  }
+
+  public async deleteSetting(name: SettingName) {
+    return this.settingsService.settings().deleteSetting(name);
+  }
+
   private constructServices() {
     this.createPayloadManager();
     this.createItemManager();
@@ -1470,6 +1492,7 @@ public getSessions(): Promise<(HttpResponse & { data: RemoteSession[] }) | HttpR
     this.createComponentManager();
     this.createActionsManager();
     this.createPreferencesService();
+    this.createSettingsService();
   }
 
   private clearServices() {
@@ -1493,6 +1516,7 @@ public getSessions(): Promise<(HttpResponse & { data: RemoteSession[] }) | HttpR
     (this.preferencesService as unknown) = undefined;
     (this.permissionsService as unknown) = undefined;
     (this.credentialService as unknown) = undefined;
+    (this.settingsService as unknown) = undefined;
 
     this.services = [];
   }
@@ -1544,7 +1568,7 @@ public getSessions(): Promise<(HttpResponse & { data: RemoteSession[] }) | HttpR
       this.httpService,
       this.storageService,
       this.permissionsService,
-      this.defaultHost,
+      this.defaultHost
     );
     this.services.push(this.apiService);
   }
@@ -1634,7 +1658,7 @@ public getSessions(): Promise<(HttpResponse & { data: RemoteSession[] }) | HttpR
       this.alertService,
       this.protocolService,
       this.challengeService,
-      this.permissionsService,
+      this.permissionsService
     );
     this.serviceObservers.push(
       this.sessionManager.addEventObserver(async (event) => {
@@ -1643,9 +1667,11 @@ public getSessions(): Promise<(HttpResponse & { data: RemoteSession[] }) | HttpR
             void (async () => {
               await this.sync();
               if (this.protocolService.needsNewRootKeyBasedItemsKey()) {
-                void this.protocolService.createNewDefaultItemsKey().then(() => {
-                  void this.sync();
-                })
+                void this.protocolService
+                  .createNewDefaultItemsKey()
+                  .then(() => {
+                    void this.sync();
+                  });
               }
             })();
             break;
@@ -1765,6 +1791,13 @@ public getSessions(): Promise<(HttpResponse & { data: RemoteSession[] }) | HttpR
       })
     );
     this.services.push(this.preferencesService);
+  }
+
+  private createSettingsService() {
+    this.settingsService = new SNSettingsService(
+      this.sessionManager,
+      this.apiService
+    );
   }
 
   private getClass<T>(base: T) {
--- a/packages/snjs/lib/services/api/api_service.ts
+++ b/packages/snjs/lib/services/api/api_service.ts
@@ -15,6 +15,11 @@ import {
   KeyParamsResponse,
   SessionListResponse,
   RawSyncResponse,
+  ListSettingsResponse,
+  UpdateSettingResponse,
+  GetSettingResponse,
+  DeleteSettingResponse,
+  MinimalHttpResponse,
 } from './responses';
 import { Session, TokenSession } from './session';
 import { ContentType } from '@Models/content_types';
@@ -48,6 +53,8 @@ type PathNamesV1 = {
   session: (sessionUuid: string) => string;
   itemRevisions: (itemId: string) => string;
   itemRevision: (itemId: string, revisionId: string) => string;
+  settings: (userUuid: string) => string;
+  setting: (userUuid: string, settingName: string) => string;
 };
 
 const Paths: {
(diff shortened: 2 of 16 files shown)


packages/snjs/lib/services/api/api_service.ts
```diff
@@ -575,6 +579,82 @@ export class SNApiService extends PureService {
...
+          errorResponse,
+          params.fallbackErrorMessage
+        );
+      });
+    this.processResponse(response);
+    return response as T;
+  }
+
+  async listSettings(userUuid: UuidString): Promise<ListSettingsResponse> {
+    const resp = await this.tokenRefreshableRequest<ListSettingsResponse>({
```
mobitar, line 606:
```suggestion
    const response = await this.tokenRefreshableRequest<ListSettingsResponse>({
```

Better for search/readability

packages/snjs/lib/services/settings_service/SNSettingsService.ts
```diff
@@ -0,0 +1,29 @@
...
+import { User } from '../api/responses';
+import { SettingsRepo } from './SettingsRepo';
+
+export class SNSettingsService extends PureService {
+  private readonly _provider: SettingsProvider;
+  private readonly getUser: () => User | null;
+
+  constructor(storageService: SNStorageService, apiService: SNApiService) {
+    super();
+    this.getUser = (): User | null =>
```
mobitar, line 16: I don't think you can do this inside the constructor because storageService may not yet be ready to read values on construction. There are stages to application initialization and storage service loading is async.

packages/snjs/lib/services/settings_service/SettingsRepo.ts
```diff
@@ -0,0 +1,100 @@
...
+    if (error != null) throw new Error(error.message);
+    if (setting?.value == null) return null;
+    return JSON.parse(setting.value);
+  }
+
+  async updateSetting<Key extends keyof SettingPayload>(
+    name: Key,
+    payload: SettingPayload[Key]
+  ): Promise<SettingPayload[Key] | null> {
+    const value = JSON.stringify(payload);
```
mobitar, line 82: I think it would be better for only the final settings repository to be JSON stringified/parsed, rather than stringifying and parsing every value. This is how storage service operates and it makes for a less confusing API.

How are settings persisted to disk? You may not even need to think about JSON if you are passing the final repository to StorageService.

packages/snjs/lib/services/settings_service/SettingsRepo.ts
```diff
@@ -0,0 +1,100 @@
...
+  ): Promise<DeleteSettingResponse>;
+}
+
+export class SettingsRepo implements SettingsProvider {
+  constructor(
+    private readonly api: SettingsAPI,
+    private readonly getUser: () => User | null
+  ) {}
+
+  canModify() {
```
mobitar, line 39: Can this be more descriptive? Perhaps

```suggestion
  isReady() {
```

or 

```suggestion
  isReadyForModification() {
```

packages/snjs/lib/application.ts
```diff
@@ -1441,6 +1446,10 @@ public getSessions(): Promise<(HttpResponse & { data: RemoteSession[] }) | HttpR
     }
   }
 
+  public get settings(): SettingsProvider {
```
mobitar, line 1449: The idea with Application is that it's the highest level interface for clients to interface with services. So it should not expose public properties like this. Instead it should expose the API of the services directly via high-level wrapper functions.

packages/snjs/lib/services/settings_service/SettingsRepo.ts
gorjan5sk: Done. Resolving comment.

packages/snjs/lib/application.ts
gorjan5sk: Done. Resolving comment.

packages/snjs/lib/services/settings_service/SettingsRepo.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  async listSettings() {
+    const { error, settings } = await this.api.listSettings(this.userUuid);
+    if (error != null) throw new Error(error.message);
+    if (settings == null) return {};
+
+    const settingVals = Object.values<string>(SettingName);
+
+    const res: Partial<SettingPayload> = {};
+    for (const s of settings) {
```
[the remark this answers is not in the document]
gorjan5sk: Done. Resolving comment.

packages/snjs/lib/services/settings_service/SNSettingsService.ts
gorjan5sk: I saw that pattern in other services, but just wanted to try if this would work - found that as you said, the services have to respect the application lifecycle :)
Fixed now. Resolving comment.

packages/snjs/lib/services/api/api_service.ts
gorjan5sk: I did a direct return without an intermediate "resp" variable which I needed for debugging. Thanks for the heads up!

packages/snjs/lib/services/settings_service/SettingsRepo.ts
gorjan5sk: Yep, you're right! It's fixed now.

The settings aren't persisted yet in this PR. Wanted to get the test going with the direct access first.

packages/snjs/lib/services/settings_service/SNSettingsService.ts
```diff
@@ -0,0 +1,31 @@
...
+import { SNApiService } from '../api/api_service';
+import { SettingsProvider } from './SettingsProvider';
+import { SettingsGateway } from './SettingsGateway';
+import { SNSessionManager } from '../api/session_manager';
+
+export class SNSettingsService extends PureService {
+  private _provider!: SettingsProvider;
+
+  constructor(
+    private readonly session: SNSessionManager,
```
mobitar, line 12:
```suggestion
    private readonly sessionManager: SNSessionManager,
```

packages/snjs/lib/services/settings_service/SNSettingsService.ts
```diff
@@ -0,0 +1,31 @@
...
+    private readonly api: SNApiService
```
mobitar, line 13:
```suggestion
    private readonly apiService: SNApiService
```

packages/snjs/lib/services/settings_service/SNSettingsService.ts
```diff
@@ -0,0 +1,31 @@
...
+    this._provider = new SettingsGateway(this.api, getUser);
+  }
+
+  settings(): SettingsProvider {
+    return this._provider;
+  }
+
+  deinit(): void {
+    (this._provider as unknown) = undefined;
+    (this.session as unknown) = undefined;
```
mobitar, line 29: Need to deinit api service

packages/snjs/lib/services/settings_service/SettingsGateway.ts
```diff
@@ -0,0 +1,97 @@
...
+}
+
+/**
+ * SettingsGateway coordinates communication with the API service
+ * and transforms settings' payloads between ready to use objects
+ * and serialized values for the backend
+ */
+export class SettingsGateway implements SettingsProvider {
+  constructor(
+    private readonly api: SettingsAPI,
```
mobitar, line 40:
```suggestion
    private readonly settinsgApi: SettingsAPI,
```

packages/snjs/lib/services/settings_service/SettingsGateway.ts
```diff
@@ -0,0 +1,97 @@
...
+    }
+    return user.uuid;
+  }
+
+  async listSettings() {
+    const { error, data } = await this.api.listSettings(this.userUuid);
+    if (error != null) throw new Error(error.message);
+    if (data == null || data.settings == null) return {};
+
+    const res: Partial<SettingPayload> = {};
```
mobitar, line 61: What does res stand for?

gorjan5sk: Changed it to settings. It's the whole list of settings.

packages/snjs/lib/services/settings_service/SettingsGateway.ts
gorjan5sk: Done in latest commit.

packages/snjs/lib/services/settings_service/SNSettingsService.ts
gorjan5sk: Done in latest commit.

packages/snjs/lib/services/settings_service/SNSettingsService.ts
gorjan5sk: Done in latest commit.

packages/snjs/lib/services/settings_service/SNSettingsService.ts
gorjan5sk: Done in latest commit.

gorjan5sk:
@mobitar I had to bring back serialization and deserialization in setting update and get, because the setting value sent to and returned from the server is a string. The tests were failing because of it.

packages/snjs/lib/services/settings_service/SNSettingsService.ts
```diff
@@ -0,0 +1,32 @@
...
+  constructor(
+    private readonly sessionManager: SNSessionManager,
+    private readonly apiService: SNApiService
+  ) {
+    super();
+  }
+
+  initializeFromDisk() {
+    const getUser = () => this.sessionManager.getUser();
+    this._provider = new SettingsGateway(this.apiService, getUser);
```
mobitar, line 20: Feel like there could be a retention cycle here. Why not pass the user directly to the gateway? Or pass the session manager?

The cycle is that `this` retains provider, provider retains `this` via the `getUser` function.

mobitar: When you release your reference in deinit in this class, the settings service will likely not be deallocated because it's still being held onto by the gateway.

packages/snjs/test/lib/settings.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    const password = Uuid.GenerateUuidSynchronously();
+    const ephemeral = false;
+    const mergeLocal = true;
+
+    await snApp.register(email, password, ephemeral, mergeLocal);
+
+    // We have to wait for sync to finish, otherwise
+    // the SNSyncService.handleSuccessServerResponse is fired
+    // after the test has finished causing an error
+    await snApp.sync();
```
mobitar, line 32: This isn't the case with other test suites. Shouldn't be necessary here.

packages/snjs/test/lib/settings.test.ts
```diff
@@ -0,0 +1,73 @@
...
+  });
+
+  afterEach(async function () {
+    await snApp!.deinit(DeinitSource.SignOut);
```
mobitar, line 36:
```suggestion
    await Factory.safeDeinit(snApp);
```

packages/snjs/test/lib/settings.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    localStorage.clear();
```
mobitar, line 37: Handled via global afterEach
```suggestion
```

packages/snjs/test/lib/settings.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    expect(responseList).toEqual({ [fakeSetting]: fakePayload });
+  });
+
+  it('creates and deletes a setting', async function () {
+    // Create and validate
+    await snApp.updateSetting(fakeSetting, fakePayload);
+    const responseCreate = await snApp.getSetting(fakeSetting);
+    expect(responseCreate).toEqual(fakePayload);
+
+    // Delete and validate
```
mobitar, line 60: Comments unnecessary

packages/snjs/test/lib/settings.test.ts
mobitar: (Likely a sign that something is incorrect)

gorjan5sk:
@mobitar addressed all the feedback in latest commit.

packages/snjs/test/lib/settings.test.ts
gorjan5sk: It was a remnant of my attempt to modify the old testing infrastructure for this purpose. There's no need now for that. I think it was the Factory.safeDeinit that solved it.

packages/snjs/lib/services/settings_service/SNSettingsService.ts
gorjan5sk: Fixed.

packages/snjs/test/lib/settings.test.ts
gorjan5sk: Fixed.

packages/snjs/test/lib/settings.test.ts
gorjan5sk: Removed.

packages/snjs/test/lib/settings.test.ts
gorjan5sk: Removed.

packages/snjs/lib/services/settings_service/SettingsGateway.ts
```diff
@@ -0,0 +1,99 @@
...
+    if (error != null) throw new Error(error.message);
+  }
+
+  async deleteSetting(name: SettingName): Promise<void> {
+    const { error } = await this.settingsApi.deleteSetting(this.userUuid, name);
+    if (error != null) throw new Error(error.message);
+  }
+
+  deinit() {
+    (this.settingsApi as unknown) = null;
```
mobitar, line 96: If it's all the same I would rather aim for consistency with rest of deinit functions, then have future developers guess which one to use. 
```suggestion
    (this.settingsApi as unknown) = undefined;
```

gorjan5sk: Ohh, sorry, force of habit. Fixed.

mobitar (approved):

PR #385, opened 2021-08-12
feat: features instead of permissions

antsgar:
- Refactored permissions service to features service
- Added event for API service that broadcasts new user roles when receiving them on the meta field of an API response
- Features service listens both for API service events and WebSockets events, and when receiving them updates roles, fetches new features and maps them to items (creates a new item for a feature if it doesn't exist, updates its content if it does and deletes it if the feature has expired)


vardan-arm (approved):

packages/features/src/Domain/Feature/Feature.ts
```diff
@@ -28,5 +28,5 @@ export type Feature = {
     borderColor: string;
   };
   deletionWarning?: string;
-  expiresAt?: number;
+  expiresAt?: Date;
```
karolsojko, line 31: hmm this should be a number in microseconds - I think everywhere where it's possible we should microsecond timestamps instead of Dates, wdyt?

karolsojko (approved):

packages/snjs/lib/services/features_service.ts
```diff
@@ -0,0 +1,181 @@
...
+          item.content.package_info
+        ) {
+          const itemIdentifier = item.content.package_info.identifier;
+          return itemIdentifier === feature.identifier && !item.deleted;
+        }
+        return false;
+      });
+      if (existingItem) {
+        if (feature.expiresAt! < today) {
+          itemsToDeleteUuids.push(existingItem.uuid);
```
mobitar, line 150: We would actually want to put this item in readonly mode when it expires: 

- https://standardnotes.com/help/20/what-happens-to-my-data-when-my-subscription-expires
- https://github.com/standardnotes/web/blob/develop/app/assets/javascripts/directives/views/componentView.ts#L207

mobitar: So I think in either condition we would just run 

```
 await this.itemManager.changeComponent(existingItem.uuid, (mutator) => {
            mutator.setContent(itemData);
          });
```

to update the item with the latest info.

packages/snjs/lib/services/features_service.ts
```diff
@@ -0,0 +1,181 @@
...
+        } else {
+          await this.itemManager.changeComponent(existingItem.uuid, (mutator) => {
+            mutator.setContent(itemData);
+          });
+        }
+      } else if (feature.expiresAt! >= today) {
```
mobitar, line 156: I think this would be just an `else` since we create the item but it would be set in readonly mode if expired.

But this wouldn't apply to themes. Only components.

packages/snjs/lib/services/features_service.ts
```diff
@@ -0,0 +1,181 @@
...
+    await this.setRoles(currentRoles);
+    await this.updateFeatures(userUuid);
+  }
+
+  private onWebSocketClose() {
+    this.webSocket = undefined;
+  }
+
+  deinit(): void {
+    this.removeApiServiceObserver?.();
```
mobitar, line 176: We should call `super()` here as well

packages/features/src/Domain/Feature/Feature.ts
antsgar: Yeah I don't recall my reasoning behind this change 😅 will revert it

packages/snjs/lib/services/features_service.ts
antsgar: > But this wouldn't apply to themes. Only components.

So components should follow the current logic and be deleted as opposed to set to read-only?

packages/snjs/lib/services/features_service.ts
antsgar: @mobitar could you please check if the logic makes sense now? Many thanks! 🙂

antsgar:
@mobitar @karolsojko should I also add an e2e test for this or are the unit tests sufficient?

packages/snjs/lib/services/features_service.ts
mobitar: Yes correct.

packages/snjs/lib/services/features_service.ts
```diff
@@ -0,0 +1,178 @@
...
+      await this.mapFeaturesToItems(
+        (featuresResponse as UserFeaturesResponse).data.features
+      );
+    }
+  }
+
+  private createItemDataForFeature(feature: Feature) {
+    return {
+      content_type: feature.contentType,
+      content: {
```
mobitar, line 98: You should use `FillItemContent` from SNJS here since there may be other required fields here.

packages/snjs/lib/services/features_service.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  private async mapFeaturesToItems(features: Feature[]): Promise<void> {
+    const currentItems = this.itemManager.getItems([
+      ContentType.Component,
+      ContentType.Theme,
+    ]);
+    const itemsToDeleteUuids = [];
+
+    const today = new Date();
```
mobitar, line 119:
```suggestion
    const now = new Date();
```

packages/snjs/lib/services/features_service.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    for (const feature of features) {
+      const expired = feature.expiresAt! < today.getTime();
+      const itemData = this.createItemDataForFeature(feature);      
+      const existingItem = currentItems.find((item) => {
+        if (
+          item.content &&
+          typeof item.content !== 'string' &&
```
mobitar, line 127: You could replace this whole conditional with `item.safeContent.package_info` I think.

mobitar (approved):

mobitar:
If the server is ready for it then e2e tests are essential.

antsgar:
> If the server is ready for it then e2e tests are essential.

@mobitar how do we create or refund subscriptions for a user to make their user role change?

mobitar:
> how do we create or refund subscriptions for a user to make their user role change?

I would in that case simulate a fake response by calling something like `apiService.processResponse({...})`. But not sure if that is the best way to do it.

mobitar (approved):

antsgar:
> I would in that case simulate a fake response by calling something like `apiService.processResponse({...})`. But not sure if that is the best way to do it.

The only thing thing is, if we do that the /features endpoint is still always going to return the same set of features. Maybe we could mock the server responses and do an e2e client-only test?

karolsojko:
> The only thing thing is, if we do that the /features endpoint is still always going to return the same set of features. Maybe we could mock the server responses and do an e2e client-only test?

Hmm we could do db fixtures to have a user that has more features enabled if that is the issue ?

antsgar:
> > The only thing thing is, if we do that the /features endpoint is still always going to return the same set of features. Maybe we could mock the server responses and do an e2e client-only test?
> 
> Hmm we could do db fixtures to have a user that has more features enabled if that is the issue ?

Yup the /features endpoint isn't deployed yet anyway, so I managed to solve it by mocking all the API responses for that endpoint 🙂

PR #388, opened 2021-08-17
feat: migrate extension key

antsgar:


packages/snjs/lib/migrations/2_7_23.ts
```diff
@@ -0,0 +1,31 @@
...
+import { Migration } from '@Lib/migrations/migration';
+import { ContentType } from '@Lib/models';
+import { SettingName } from '@standardnotes/settings';
+
+export class Migration2_7_23 extends Migration {
+  static version(): string {
+    return '2.7.23';
+  }
+
+  protected registerStageHandlers(): void {
```
mobitar, line 11: Sign in event is only on explicit sign in. And actually you don’t have to be signed in to have a repo. This should be on full sync completed instead.

mobitar:
Is this testable?

packages/snjs/lib/migrations/2_7_23.ts
antsgar: But we need a user uuid to be able to update the user setting

mobitar: Ah ok, I’m that case on full sync completed and if signed in.

antsgar:
> Is this testable?

It should be, let me give it a try 🙂

antsgar:
@mobitar added the extension repo items observer, but I suppose we should still keep the migration for already existing items?

antsgar:
Depends on #385 since the observer has been added to the new features service

karolsojko (approved):
Good stuff - glad to know we have these kind of mechanisms in place 👍

antsgar:
Closing in favor of #394 (PR shows wrong commits after dependent PR was squash-merged)

PR #390, opened 2021-08-20
Feature/features instead of permissions proposal

mobitar:


vardan-arm (approved):

packages/snjs/lib/services/features_service.ts
```diff
@@ -49,9 +50,9 @@ export class SNFeaturesService extends PureService<void> {
       async (eventName, data) => {
         if (eventName === WebSocketsServiceEvent.UserRoleMessageReceived) {
           const {
-            payload: { userUuid, currentRoles },
-          } = data as UserRolesChangedEvent;
-          await this.setRoles(currentRoles);
+            payload: { userUuid, toRole },
```
antsgar, line 53: This event's payload includes `currentRoles` field https://github.com/standardnotes/snjs/blob/master/packages/domain-events/src/Domain/Event/UserRolesChangedEventPayload.ts#L6. Used to be `toRole` so possibly you had the outdated dependency installed

antsgar (approved):

PR #395, opened 2021-08-25
feat: check if latest develop is compatible with snjs

karolsojko:


karolsojko:
@antsgar I need your help on this one.

I've tried to check if SNJS is working on latest auth and ssjs.

One test fails with the features service update roles observer: https://github.com/standardnotes/snjs/blob/master/packages/snjs/lib/services/features_service.ts#L45-L48

if you run only this test: https://github.com/standardnotes/snjs/blob/master/packages/snjs/test/lib/session.test.js#L105 - it passes

if you run only these 2 tests: https://github.com/standardnotes/snjs/blob/master/packages/snjs/test/lib/session.test.js#L90-L112

the second one fails. Seems like the item manager is undefined after the first test passes and second runs.

Could you have check it out?

antsgar:
@karolsojko sure I'll check it out!

antsgar:
@karolsojko is what you described happening when you run the tests locally? I can't seem to be able to reproduce it

antsgar:
Nevermind, I could reproduce now, will take a look 🙂

PR #400, opened 2021-08-31
feat: send snjs and environment versions in API call headers

vardan-arm:


packages/snjs/lib/services/api/http_service.ts
```diff
@@ -89,6 +90,21 @@ export class SNHttpService extends PureService {
...
+    if (environmentWithVersion) {
+      const [environment, version] = environmentWithVersion.split('::');
+      if (environment && version) {
+        // TODO: move this if/else below to some util; also the separator above - `::`
+        if (environment === 'web') {
+          request.setRequestHeader('X-Web-Version', version)
+        }
+      }
+    }
+    const snjsVersion = packageJson.version;
```
mobitar, line 105: This value is already available globally.

packages/snjs/lib/services/api/http_service.ts
```diff
@@ -89,6 +90,21 @@ export class SNHttpService extends PureService {
     }
     request.open(httpRequest.verb, httpRequest.url, true);
     request.setRequestHeader('Content-type', 'application/json');
+
+    // TODO: `localStorage` might not be present in mobile, perhaps need to use the service instead
+    const environmentWithVersion = localStorage.getItem('environmentWithVersion');
```
mobitar, line 95: Never access `localStorage` directly from SNJS. SNJS is _not_ a browser-centric library.

mobitar: Versions should be passed into this service's constructors as to avoid passing in another service to query the needed values.

vardan-arm: I see, I was intending to set version to storage in web/mobile/desktop and get its value here like `deviceInterface.getRawStorageValue(key)`, so in case of mobile it would be `AsyncStorage`, for example. So the implementation would have been:
```
if (typeof localStorage !== 'undefined') {
  // browser or desktop - get the version from local storage
} else if (typeof AsyncStorage !== 'undefined') {
  // mobile - get the version from async storage
}
```

> Versions should be passed into this service's constructors
Makes sense, will update that way.

packages/snjs/lib/application.ts
```diff
@@ -340,6 +346,11 @@ export class SNApplication {
     // optional override
   }
 
+  public getAppVersion(): string {
+    // mandatory override
+    return '';
```
mobitar, line 351: Should throw exception if mandatory override.
```suggestion
    throw Error('getAppVersion must be overriden.');
```

packages/snjs/lib/application.ts
```diff
@@ -1558,6 +1569,7 @@ export class SNApplication {
     (this.webSocketsService as unknown) = undefined;
     (this.settingsService as unknown) = undefined;
     (this.mfaService as unknown) = undefined;
+    (this.appVersion as unknown) = undefined;
```
mobitar, line 1572: String value doesn't need to be de-inited manually. Only complex services which hang on to other objects.

packages/snjs/lib/application.ts
vardan-arm: I actually removed this method at all since I'm already passing version in constructor, just didn't push yet. But really good approach for such cases :)

packages/snjs/lib/services/api/http_service.ts
```diff
@@ -98,6 +122,19 @@ export class SNHttpService extends PureService {
     return request;
   }
 
+  private getCorrectHeaderForEnvironment(environment: Environment): string {
```
mobitar, line 125:
```suggestion
  private headerForEnvironment(environment: Environment): string {
```

packages/snjs/lib/application.ts
vardan-arm: Noted, done 👍

packages/snjs/lib/services/api/http_service.ts
```diff
@@ -26,10 +28,27 @@ export type HttpRequest = {
...
+
 /**
  * A non-SNJS specific wrapper for XMLHttpRequests
  */
 export class SNHttpService extends PureService {
+  private readonly appVersion: string;
+  private readonly environment: Environment;
+
+  constructor({
+    environment,
```
mobitar, line 44:
```suggestion
    private readonly environment: string,
```

This is shorthand for property notation in TS, and removes the need for declaring definitions on line 40 and setting values on line 49.

packages/snjs/lib/services/api/http_service.ts
mobitar:
```suggestion
  private versionHeaderName(): string {
```

This function also doesn't need to take a param since this.environment is already available to it.

vardan-arm: I'm usually prepending `get` to such method names so that they differ from property names. Updated. :)

packages/snjs/lib/services/api/http_service.ts
vardan-arm: I like this approach when arguments are provided one by one, but in our case constructor receives an object as an argument that contains `environment` and `appVersion`. Seems there isn't a good way to use shorthand notation for such cases.

mobitar: Object approach in service constructors would be the exception in our design. Probably no sense in modifying the pattern now. What's the benefit to objects in constructor if using TS? Usually they made sense in JS because you could make sure you're passing the right values regardless of order, but TS solves that problem.

vardan-arm: If the arguments count is small, then passing arguments one by one doesn't give problems. But if there are many (e.g., 4+) arguments, using objects does help. It's also very useful if some of those arguments are optional, since in non-object approach if you add new mandatory argument you need to either put it before the optional one or add empty value for optional argument.
But yeah, in this case we don't have any such case, so there is no any benefit from using object, so I updated it :)

vardan-arm:
Internal link: https://app.asana.com/0/1200588672119490/1200383926936362

packages/snjs/lib/services/api/http_service.ts
```diff
@@ -30,6 +32,15 @@ export type HttpRequest = {
  * A non-SNJS specific wrapper for XMLHttpRequests
  */
 export class SNHttpService extends PureService {
+  constructor(
+    private readonly environment: Environment,
+    private readonly appVersion: string
+  ) {
+    super();
+
+    this.appVersion = appVersion;
```
mobitar, line 41: This can be removed now, because the TS shorthand notation automatically does this :)

vardan-arm: Ah, sure, forgot removing them, thanks! :)

packages/snjs/lib/services/api/http_service.ts
```diff
@@ -98,6 +111,19 @@ export class SNHttpService extends PureService {
...
+        return 'X-Web-Version';
+      case Environment.Mobile:
+        return 'X-Mobile-Version';
+      case Environment.Desktop:
+        return 'X-Desktop-Version';
+      default:
+        throw Error(`'${this.environment}' is not correct environment value`);
+    }
+  }
+
```
gorjan5sk, line 126: I'm not sure on the backend requirements, @karolsojko correct me if I'm wrong, but wouldn't it be better to have a single header name, and the header value would contain something like `Web: x.y.z`?

Searching/reporting from the logs for stuff wouldn't require us to have a complex query string like `"X-Web" OR "X-Mobile"...

karolsojko: it's common for a lot of servers to ditch custom headers that are not `X-...` prefixed. Event though there's an RFC of deprecating the `X-...` naming convention - if we want to add custom defined headers we have to keep them in the `X-...` to be compatible with what's out there I think

gorjan5sk: I think I didn't explain myself. I was proposing of having a single header name instead of three different header names.

karolsojko: hmm let's say we have web at version 1.2.3 and mobile at version 1.2.3 how would we know if the request is coming from a mobile client or web client?

gorjan5sk: `X-Version: Web-1.2.3` vs `X-Version: Android-1.2.3`?

karolsojko: Good idea. @vardan-arm so let's have `X-SNJS-Version` and `X-Application-Version`

vardan-arm: Good point @gorjan5sk , updated.

packages/snjs/lib/application.ts
```diff
@@ -183,6 +183,7 @@ export class SNApplication {
    * and 'with' is the custom subclass to use.
    * @param skipClasses An array of classes to skip making services for.
    * @param defaultHost Default host to use in ApiService.
+   * @param appVersion Version of application where SNApplication is instantiated.
```
mobitar, line 186:
```suggestion
   * @param appVersion Version of client application.
```

mobitar (approved):

vardan-arm:
Does it make sense to do major version bump for these changes, since it contains breaking changes (all clients must provide application version with these changes and will break otherwise)?

PR #401, opened 2021-09-01
feat: add ContentType to common package

karolsojko:
Internal link: https://app.asana.com/0/1160985268757881/1200878677649176/f


vardan-arm:
Although it's already merged, but should we keep https://github.com/standardnotes/snjs/blob/master/packages/features/src/Domain/Content/ContentType.ts after these changes?

vardan-arm:
Ah, got it, yeah, it's covered in that new PR :)

PR #403, opened 2021-09-01
feat: store user features

antsgar:
Store user features fetched from /features endpoint.

### DIFF
--- a/packages/snjs/lib/services/features_service.ts
+++ b/packages/snjs/lib/services/features_service.ts
@@ -89,10 +89,16 @@ export class SNFeaturesService extends PureService<void> {
   }
 
   public async loadUserRolesAndFeatures(): Promise<void> {
-    this.roles =
-      (await this.storageService.getValue(StorageKey.UserRoles)) || [];
-    this.features =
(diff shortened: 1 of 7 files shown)


packages/snjs/lib/services/features_service.ts
```diff
@@ -86,9 +88,11 @@ export class SNFeaturesService extends PureService<void> {
     }
   }
 
-  public async loadUserRoles(): Promise<void> {
+  public async loadUserRolesAndFeatures(): Promise<void> {
     this.roles =
       (await this.storageService.getValue(StorageKey.UserRoles)) || [];
+    this.features =
+      (await this.storageService.getValue(StorageKey.UserFeatures)) || [];
```
mobitar, line 95: I believe getValue has a defaultValue param, if it makes things easier.

mobitar (approved):

packages/snjs/lib/services/features_service.ts
antsgar: Good catch, I've updated it 🙂

PR #407, opened 2021-09-02
feat: add USER_CHANGED_EMAIL event

karolsojko:


vardan-arm (approved):

packages/domain-events/src/Domain/Event/UserChangedEmailEventPayload.ts
```diff
@@ -0,0 +1,5 @@
+export interface UserChangedEmailEventPayload {
```
mobitar, line 1: Nitpick but the pattern with other events is that the verb is last. So this would be `UserEmailChanged` event.

karolsojko: Fixed in: https://github.com/standardnotes/snjs/pull/409

PR #411, opened 2021-09-06
feat: add get user subscription api call

antsgar:
https://app.asana.com/0/1199914526147392/1200845194469719/f


mobitar (approved):

vardan-arm (approved):

vardan-arm:
@antsgar as an info `@standardnotes/snjs` with version `2.1.2` is already merged by [this](https://github.com/standardnotes/snjs/pull/417) PR, so I think it makes sense to bump it to `2.1.3` here.

packages/snjs/lib/application.ts
```diff
@@ -554,6 +554,14 @@ export class SNApplication {
...
   }
 
+  public getUserSubscription(): Promise<HttpResponse | GetSubscriptionResponse> {
+    return this.sessionManager.getSubscription();
+  }
+
+  public getSubscriptions(): Promise<HttpResponse | GetSubscriptionsResponse> {
+    return this.apiService.getSubscriptions();
+  }
+
```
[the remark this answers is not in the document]
mobitar: Agreed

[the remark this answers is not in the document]
antsgar: Isn't it possible that we could want for example the error message from the error response to get to the frontend? Also checking, it seems we're already using this pattern on the sessions methods exposed on application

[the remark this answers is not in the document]
antsgar: I've changed it 🙂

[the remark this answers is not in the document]
gorjan5sk: > Isn't it possible that we could want for example the error message from the error response to get to the frontend? Also checking, it seems we're already using this pattern on the sessions methods exposed on application

The common pattern is that we put the error in the Promise. Unfortunately, Typescript doesn't allow type inference for Errors in the Promises so we have to use the base form. But that's fine in this case since you don't want to process statuses in the view layer, you just trigger events and show the message once processed. The drawback of that approach comes when you want to support multiple languages, but the error messages are the least of the concerns since they're easy to discover - just search for ".catch" and "Error(.*)".

packages/snjs/lib/application.ts
```diff
@@ -559,6 +560,28 @@ export class SNApplication {
...
 
+  public async getUserSubscription(): Promise<Subscription | undefined> {
+    try {
+      const response = await this.sessionManager.getSubscription();
+      if (!response.error && response.data) {
+        return (response as GetSubscriptionResponse).data!.subscription
+      }
+    } catch (e) {
+      return undefined;
+    }
```
gorjan5sk, line 571: We want the errors to propagate:
```
if(response.error != null) throw new Error(error.message);
```

gorjan5sk: And no need to wrap in try/catch. The error that happens in await will automatically be embedded in the error response part of the Promise.

packages/snjs/lib/application.ts
```diff
@@ -559,6 +560,28 @@ export class SNApplication {
...
+  public async getAvailableSubscriptions(): Promise<AvailableSubscriptions | undefined> {
+    try {
+      const response = await this.apiService.getAvailableSubscriptions();
+      if (!response.error && response.data) {
+        return (response as GetAvailableSubscriptionsResponse).data!
+      }
+    } catch (e) {
+      return undefined;
+    }
+  }
```
gorjan5sk, line 583: Same here as above.

packages/snjs/lib/application.ts
antsgar: Done!

karolsojko (approved):

packages/snjs/lib/services/api/api_service.ts
```diff
@@ -700,6 +712,31 @@ export class SNApiService extends PureService<
...
+    userUuid: string
+  ): Promise<HttpResponse | GetSubscriptionResponse> {
+    const url = joinPaths(this.host, Paths.v1.subscription(userUuid));
+    const response = await this.request({
+      verb: HttpVerb.Get,
+      url,
+      authentication: this.session?.authorizationValue,
+      fallbackErrorMessage: messages.API_MESSAGE_FAILED_SUBSCRIPTION_INFO,
+    });
+    this.processResponse(response);
```
gorjan5sk, line 725: I think processResponse already happens in `this.request`.

gorjan5sk (approved):
Approved! :) One small comment regarding processResponse.

packages/snjs/lib/services/api/api_service.ts
antsgar: Ah good catch, thanks! 🙂

packages/snjs/lib/application.ts
```diff
@@ -559,6 +560,28 @@ export class SNApplication {
     return compareVersions(userVersion, ProtocolVersion.V004) >= 0;
   }
 
+  public async getUserSubscription(): Promise<Subscription | undefined> {
+    const response = await this.sessionManager.getSubscription();
+    if (response.error) {
+      throw new Error(response.error.message)
```
mobitar, line 566: Hmm why are we throwing? We should propagate this to the client so the user can take action. Otherwise this might block execution of other essential tasks.

gorjan5sk: When throwing an error in an async function it triggers the catch/error of the Promise. That's how it's propagated towards the client.

antsgar:
@mobitar the failing e2e test here also failed in your latest commit in the master branch. Safe to ignore?

mobitar:
Yep need to look into it. Is a test/deinit issue rather than core issue.

PR #417, opened 2021-09-08
feat: add function to return password creation date for user

vardan-arm:


vardan-arm:
Internal link - https://app.asana.com/0/1200588672119490/1200888182977784

karolsojko (approved):

packages/snjs/lib/services/api/session_manager.ts
```diff
@@ -120,6 +120,11 @@ export class SNSessionManager extends PureService<SessionEvent> {
     }
   }
 
+  public getPasswordCreatedDate(): number | undefined {
+    const rootKey = this.protocolService.getRootKey()
```
mobitar, line 124:
```suggestion
    const rootKey = this.protocolService.getRootKey();
```

packages/snjs/lib/services/api/session_manager.ts
```diff
@@ -120,6 +120,11 @@ export class SNSessionManager extends PureService<SessionEvent> {
...
+    return rootKey ? Number(rootKey.keyParams.content.created) : undefined;
```
mobitar, line 125:
```suggestion
    return rootKey ? rootKey.keyParams.createdDate : undefined;
```

mobitar: This would change the return type to date. Do you need it to be a Number?

vardan-arm: No, I just didn't want stringified timestamp, therefore converted to Number :) Thanks, updated!

packages/snjs/lib/services/api/session_manager.ts
vardan-arm: Done

packages/snjs/lib/services/api/session_manager.ts
```diff
@@ -120,6 +120,11 @@ export class SNSessionManager extends PureService<SessionEvent> {
     }
   }
 
+  public getPasswordCreatedDate(): Date | undefined {
```
mobitar, line 123: Sorry didn't catch this earlier, but I would put this function in protocol_service instead since root key is the domain of the protocol service.

mobitar (approved):

packages/snjs/lib/services/api/session_manager.ts
vardan-arm: Good point, yeah, `getRootKey` comes from protocol service, so it makes more sense, done.

PR #429, opened 2021-09-13
feat: make disabling mfa a protected action

mobitar:


packages/snjs/lib/services/protection_service.ts
```diff
@@ -274,6 +280,12 @@ export class SNProtectionService extends PureService<ProtectionEvent.SessionExpi
     if (this.protocolService.hasPasscode()) {
       prompts.push(new ChallengePrompt(ChallengeValidation.LocalPasscode));
     }
+    if (requireAccountPassword) {
+      if (!this.protocolService.hasAccount()) {
+        throw Error('Requiring account password for challenge with no account');
+      }
+      prompts.push(new ChallengePrompt(ChallengeValidation.AccountPassword));
```
antsgar, line 287: Will this prompt for both the passcode and the password, if the user has a passcode set?

antsgar (approved):

packages/snjs/lib/services/protection_service.ts
mobitar: Correct, we want to prompt for both.

PR #430, opened 2021-09-14
feat: use observer instead of migration to set extension key user setting

antsgar:


packages/snjs/lib/services/features_service.ts
```diff
@@ -73,9 +77,20 @@ export class SNFeaturesService extends PureService<void> {
...
       async (changed, inserted) => {
         const items = [...changed, ...inserted];
-        await this.updateExtensionKeySetting(items);
+        if (this.sessionManager.getUser()) {
+          await this.updateExtensionKeySetting(items);
+        }
       }
     );
+
+    this.removeFullSyncCompletedObserver = this.syncService.addEventObserver(
```
antsgar, line 86: @mobitar had to observe the FullSyncCompleted event since LocalDataLoaded was firing after the application was launched, but before sign in. Is this okay?

packages/snjs/lib/services/features_service.ts
```diff
@@ -73,9 +77,20 @@ export class SNFeaturesService extends PureService<void> {
...
+      async (eventName) => {
+        if (eventName === SyncEvent.FullSyncCompleted && this.sessionManager.getUser()) {
+          const extensionRepoItems = this.itemManager.getItems(ContentType.ExtensionRepo);
+          await this.updateExtensionKeySetting(extensionRepoItems);
+        }
+      }
+    )
   }
 
   public async updateExtensionKeySetting(extensionRepoItems: SNItem[]): Promise<void> {
```
antsgar, line 96: @mobitar not sure if the logic is okay here, I'm iterating over the extensionRepoItems but there's only one extension key user setting. Safe to assume there will only be one extension repo item? Should I filter them somehow?

packages/snjs/lib/services/features_service.ts
mobitar: I don't think this is the way to go, especially because you'd be running this logic on every single sync, when it should really be run once in a blue moon.

When you sign in after application launches, the `his.itemManager.addObserver` covers this use case and you'll get notified.

mobitar: That is, after sign in, all ExtensionRepo items will appear in the `inserted` part of that observer.

packages/snjs/lib/services/features_service.ts
mobitar: Hmm, users can have custom repos. In this case we only want to migrate our official repo. But I suppose it's safe to assume only we use that regex format in our URL, so we can keep as is.

packages/snjs/mocha/features.test.js
```diff
@@ -226,4 +230,29 @@ describe('features', () => {
...
+
+  describe('full sync completed observer', () => {
+    it('should update extension key user setting when local data is loaded and user is signed in', async () => {
+      expect(await application.getSensitiveSetting(SettingName.ExtensionKey)).to.equal(false);
+      const extensionKey = Uuid.GenerateUuidSynchronously().split('-').join('');
+      application = await Factory.signOutApplicationAndReturnNew(application);
+      sinon.stub(application.itemManager, 'getItems').callsFake(() => {
+        return [
+          {
+            safeContent: {
```
mobitar, line 242: safeContent is a getter, not an actual property.

packages/snjs/mocha/features.test.js
```diff
@@ -226,4 +230,29 @@ describe('features', () => {
...
   });
+
+  describe('full sync completed observer', () => {
+    it('should update extension key user setting when local data is loaded and user is signed in', async () => {
+      expect(await application.getSensitiveSetting(SettingName.ExtensionKey)).to.equal(false);
+      const extensionKey = Uuid.GenerateUuidSynchronously().split('-').join('');
+      application = await Factory.signOutApplicationAndReturnNew(application);
+      sinon.stub(application.itemManager, 'getItems').callsFake(() => {
+        return [
+          {
```
mobitar, line 241: The object being returned here is not really an item. It's missing content_type. So how does the ExtensionRepo observer in features_service pick this up?

antsgar: This replaces the actual getItems method with a method that always returns this, the idea is just to mock a situation where there's a pre-existing extension repo item

mobitar: Right but the observers are notified based on the content_type of the item

packages/snjs/mocha/features.test.js
antsgar: Yup I know but this is just a mock function, can we mock a getter in a different way?

packages/snjs/mocha/features.test.js
antsgar: I've now refactored it so that it takes contentType into account to make it more similar to the actual method

packages/snjs/mocha/features.test.js
mobitar: Hmm, I suppose it's just a weird way to write the test. You're relying too much on the fact that the client is calling safeContent, but its use is optional in the code.

packages/snjs/mocha/features.test.js
mobitar: So it was working without it? Weird.

antsgar: > Right but the observers are notified based on the content_type of the item

That's for the itemManager observer, which is the test right above this one and doesn't use a mock. This one is for the FullSyncCompleted observer

packages/snjs/lib/services/features_service.ts
antsgar: Ahhh so we'd just need the itemManager observer? That's much more simple than I thought, I guess all I needed to do then was remove the migration

mobitar: I think actually we may not need reading local db state. I forget if `inserted` includes items on first load, but you should be able to double check that. If it doesn't, then yes all you need is the observer. If `inserted` doesn't include items on first load, then you would need the observer + reading initial state.

antsgar: I added a test to try this, and `inserted` does seem to include items on first load, but at this point the user isn't yet logged in and the user setting can't be updated. `changed` does fire after login, but more than once, which would mean updating the setting more than once -- not ideal. Not sure how to solve this, since I can't use the LocalDataLoaded event either because it also fires before sign in

mobitar: You’ll want to delete the Repo item after it’s been migrated. Also read the source argument of the observer callback and then only act on a particular source you’re interested in to prevent duplicate calls.

antsgar: New PR here #431, added the item deletion but not sure if the filtered sources are the correct ones

PR #431, opened 2021-09-14
fix: leave only item manager observer

antsgar:


packages/snjs/lib/services/features_service.ts
```diff
@@ -72,8 +72,9 @@ export class SNFeaturesService extends PureService<void> {
 
     this.removeExtensionRepoItemsObserver = this.itemManager.addObserver(
       ContentType.ExtensionRepo,
-      async (changed, inserted) => {
-        if (this.sessionManager.getUser()) {
+      async (changed, inserted, _discarded, _ignored, source) => {
+        const sources = [PayloadSource.Constructor, PayloadSource.LocalChanged];
```
mobitar, line 76: I think maybe Constructor and RemoteRetrieved would be what you want. Constructor would cover initial DB load event, and remote retrieved when it is retrieved from the server.

mobitar: But I’m not sure Constructor is what we want. Maybe LocalRetrieved but I have to double check what that is

mobitar: Yup LocalRetrieved is what covers initial DB load items: https://github.com/standardnotes/snjs/blob/9133cefec991f32a29511dafb2eef094dc598244/packages/snjs/lib/services/sync/sync_service.ts#L295

packages/snjs/lib/services/features_service.ts
```diff
@@ -87,6 +91,7 @@ export class SNFeaturesService extends PureService<void> {
           const userKey = userKeyMatch[0];
           await this.settingsService
             .updateSetting(SettingName.ExtensionKey, userKey, true);
+          await this.itemManager.setItemToBeDeleted(item.uuid);
```
antsgar, line 94: @mobitar wait I've just realized, don't we need the extension repo items for our current extension manager logic?

packages/snjs/lib/services/features_service.ts
antsgar: Tried changing it to using those two sources, but both tests fail 😕

packages/snjs/lib/services/features_service.ts
mobitar: We do. We can either coordinate the release of this with the removal of ext-manager, or we can modify the Repo object to say `content.migratedToUserSetting = true` and sync it.

packages/snjs/lib/services/features_service.ts
mobitar: Let me try running this locally..

mobitar (approved):
Handing it back over to you :)

antsgar:
@mobitar awesome, thank you very much!

PR #442, opened 2021-09-30
fix: handle server rejected payloads

mobitar:


packages/snjs/lib/protocol/payloads/deltas/remote_rejected.ts
```diff
@@ -0,0 +1,33 @@
+import { CreateSourcedPayloadFromObject } from '@Payloads/generator';
+import { PayloadsDelta } from '@Payloads/deltas/delta';
+import { PayloadSource } from '@Payloads/sources';
+import { ImmutablePayloadCollection } from '@Protocol/collection/payload_collection';
+
+export class DeltaRemoteRejected extends PayloadsDelta {
+  // eslint-disable-next-line @typescript-eslint/require-await
+  public async resultingCollection(): Promise<ImmutablePayloadCollection> {
```
antsgar, line 8: Why do we want this to be async?

antsgar (approved):

packages/snjs/lib/protocol/payloads/deltas/remote_rejected.ts
mobitar: It's overriding parent function which is async. Removing async in subclass implementation results in a warning. So only way was to add the eslint-disable rule.

PR #447, opened 2021-10-06
fix: more strict component mapping

atmoio:
- Fixes issue where mapping could occur before account data has been downloaded
- Introduces more strict mapping behavior

### DIFF
--- a/packages/snjs/lib/models/core/item.ts
+++ b/packages/snjs/lib/models/core/item.ts
@@ -476,7 +476,7 @@ export class ItemMutator {
   }
 
   /** Not recommended to use as this might break item schema if used incorrectly */
-  public unsafe_setCustomContent(content: PayloadContent) {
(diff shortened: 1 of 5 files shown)


packages/snjs/lib/services/features_service.ts
```diff
@@ -239,58 +243,33 @@ export class SNFeaturesService extends PureService<void> {
...
+        resultingItem = await this.itemManager.changeComponent(
+          existingItem.uuid,
+          (mutator) => {
+            mutator.hosted_url = feature.url;
+            mutator.package_info = feature as ComponentPackageInfo;
+            mutator.valid_until = new Date(feature.expires_at || 0);
           }
-          break;
+        );
+      } else if (!expired || feature.content_type === ContentType.Component) {
```
antsgar, line 257: If I'm following correctly with this logic we won't be creating items for inexistent themes, we'd probably need to remove the second condition here?

PR #457, opened 2021-10-15
feat: Add new pref keys and implement includeProtected notes display criteria 

amanharwara:
1. Adds `NotesShowDeleted` and `NotesHideProtected` pref keys
2. Implements `includeProtected` notes display criteria to show/hide protected notes.
3. Adds a test for the above.

### DIFF
--- a/packages/snjs/package.json
+++ b/packages/snjs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@standardnotes/snjs",
-  "version": "2.14.14",
+  "version": "2.15.0",
   "engines": {
     "node": ">=14.0.0 <16.0.0"
   },


packages/snjs/lib/models/app/userPrefs.ts
```diff
@@ -31,7 +33,9 @@ export type PrefValue = {
   [PrefKey.SortNotesBy]: CollectionSort;
   [PrefKey.SortNotesReverse]: boolean;
   [PrefKey.NotesShowArchived]: boolean;
+  [PrefKey.NotesShowDeleted]: boolean;
```
mobitar, line 36: Hmm looks like we already have something similar? https://github.com/standardnotes/snjs/pull/457/files#diff-bbe8e7705d98fb9ec1181e74bc0c8bb20e12c3ac1b8596494c6c56a696de40fcR101

mobitar: Line 101
Screen Shot 2021-10-17 at 10 11 53 AM

mobitar: Ok, let's rename this to Show Trashed at the very least, as Deleted is a separate concept that is permanent and irrecoverable.

amanharwara: The design says "Show deleted notes". I guess I should change that as well?

amanharwara:
Version numbers follow semver right?

PR #461, opened 2021-10-21
feat: Add getNewSubscriptionToken() function

amanharwara:


packages/snjs/lib/services/api/api_service.ts
```diff
@@ -752,6 +752,14 @@ export class SNApiService extends PureService<
     });
     if (response.data) {
       const subscriptionToken = (response as PostSubscriptionTokensResponse).data!.token;
+      return subscriptionToken;
+    }
+    return undefined;
```
mobitar, line 757:
```suggestion
    return (response as PostSubscriptionTokensResponse).data?.token;
```

PR #474, opened 2021-11-10
fix: persist host after signing in

mobitar:


packages/snjs/lib/services/api/session_manager.ts
```diff
@@ -640,6 +640,7 @@ export class SNSessionManager extends PureService<SessionEvent> {
     const user = data.user;
     this.user = user;
     await this.storageService.setValue(StorageKey.User, user);
+    this.apiService.setHost(this.apiService.getHost());
```
johnny243, line 643: Why is this needed?

mobitar: See here https://app.asana.com/0/1200980603427013/1201146018036636

PR #483, opened 2021-11-22
feat: hide note contents if the protected note is open and the protection expires and the user didn't edit the note in previous 30 seconds

vardan-arm:
Internal link: https://app.asana.com/0/1200588672119490/1201382347168573


packages/snjs/lib/services/component_manager.ts
```diff
@@ -1053,6 +1058,10 @@ export class SNComponentManager extends PureService {
             lockedNoteCount++;
           }
         }
+        // Push only the items whose type is `SNNote`
+        if (item.protected && Object.getPrototypeOf(item) === SNNote.prototype) {
```
moughxyz, line 1062:
```suggestion
        if (item.protected && item.content_type === ContentType.Note) {
```

moughxyz (approved):

PR #484, opened 2021-11-23
WIP Folders

laurentsenta:


packages/snjs/lib/application.ts
```diff
@@ -864,21 +864,50 @@ export class SNApplication {
...
+
+  /**
+   * Remove the tag parent.
+   */
+  public async unsetTagRelationship(
+    parentTag: SNTag,
+    childTag: SNTag
+  ): Promise<void> {
+    await this.itemManager.unsetTagRelationship(parentTag, childTag);
+  }
```
laurentsenta, line 884: @moughxyz I wanted to add the "unset" operation, but I couldn't find a good antonym for `establishRelationship`, so I used `set` and `unset`, let me know if that works for you.

Note we could ignore `parentTag` in there, because in the current implementation we don't use it, I kept it anyway because:
- we can make the API symmetric with the set operation,
- we keep the option to change our internal representation later without breaking client code.

packages/snjs/lib/models/app/tag.ts
```diff
@@ -51,7 +64,22 @@ export class SNTag extends SNItem {
 }
 
 export class TagMutator extends ItemMutator {
+  get typedContent(): TagContent {
+    return this.content as TagContent;
+  }
+
   set title(title: string) {
-    this.content!.title = title;
+    this.typedContent.title = title;
```
laurentsenta, line 72: @moughxyz this is the side effect I was talking about,
we rely on a "distant side effect", someone could define:

```
get typedContent(): TagContent {
  return {...this.content} as TagContent; // or something more relevant
}
```

and break this code without breaking the interfaces.

PR #502, opened 2021-12-16
fix: use cached features until server request is made

moughxyz:


packages/snjs/lib/services/features_service.spec.ts
```diff
@@ -587,7 +593,8 @@ describe('featuresService', () => {
         featuresService.getFeatureStatus(FeatureIdentifier.TokenVaultEditor)
       ).toBe(FeatureStatus.Entitled);
 
-      featuresService['completedSuccessfulFeaturesRetrieval'] = true;
+
+      await featuresService.didDownloadFeatures(features);
```
laurentsenta, line 597: I guess we used `featuresService['completedSuccessfulFeaturesRetrieval']` to access a private method from the unit test,

Could you use the same pattern and defend the service's public API?

`featuresService['didDownloadFeatures'](features);`

PR #532, opened 2021-12-29
fix: git sync race conditions

karolsojko:
## Description

Git sync should be run:
- asynchronously on any given branch except `master` and `beta/*` branches.
- synchronously (after publishing and versioning of packages) on `master` and `beta/*` branches.

## Related Issue(s)

https://app.asana.com/0/1200980603427038/1201564753368422/f

## Checklist

- [x] This PR has NO tests

### DIFF
--- a/.github/workflows/git-sync.yml
+++ b/.github/workflows/git-sync.yml
@@ -1,6 +1,10 @@
 name: Git Sync
 
 on: push
+  branches-ignore:
+    - beta/*
+    - master
+
 jobs:
   git-sync:
     runs-on: ubuntu-latest
--- a/.github/workflows/git-sync.yml
+++ b/.github/workflows/git-sync.yml
@@ -1,9 +1,10 @@
 name: Git Sync
 
-on: push
-  branches-ignore:
-    - beta/*
-    - master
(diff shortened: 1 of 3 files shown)


.github/workflows/git-sync.yml
```diff
@@ -1,6 +1,11 @@
 name: Git Sync
 
-on: push
+on:
+  push:
+    branches-ignore:
+      - beta/*
+      - master
+
```
laurentsenta, line 8: From what I understand:

We're still pushing `"refs/remotes/source/*"` in both cases,
which means that the race condition (2 git sync running at the same time will step on each other toes) still happens,

Shouldn't we push `refs/remote/source/${CURRENT_CI_BRANCH}` so that each CI run takes care of its own branch?

karolsojko: Unfortunately that wouldn't solve the problem in case of `master` and `beta/*` branches since those branches will be affected by commits and pushes by `standardci` user. So as you can see we have to have sequential syncs in case of `master` and `beta/*` (after `standardci` changes) and async syncs on any other branch

PR #546, opened 2022-01-11
feat: tag notes index

moughxyz:


packages/snjs/lib/protocol/collection/tag_notes_index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  private recieveTagChanges(tags: SNTag[]): void {
+    for (const tag of tags) {
+      const uuids = tag.noteReferences.map((ref) => ref.uuid);
+      const countableUuids = uuids.filter((uuid) =>
+        this.allCountableNotes.has(uuid)
+      );
+      this.tagToNotesMap[tag.uuid] = new Set(countableUuids);
+    }
+  }
```
laurentsenta, line 45: a few typos, you could search and replace recieve -> receive

packages/snjs/lib/services/item_manager.ts
```diff
@@ -313,6 +324,13 @@ export class ItemManager extends PureService {
       this.collection.discard(item);
     }
     this.notesView.setNeedsRebuilding();
+    this.tagNotesIndex.receiveTagAndNoteChanges(
+      changedOrInserted.filter(
+        (item) =>
+          item.content_type === ContentType.Tag ||
+          item.content_type === ContentType.Note
+      ) as SNTag[]
+    );
```
laurentsenta, line 333: It's probably `as SNTag | SNNote` here.

packages/snjs/lib/protocol/collection/tag_notes_index.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  public countableNotesForTag(tag: SNTag): number {
+    return this.tagToNotesMap[tag.uuid]?.size;
+  }
+
+  public receiveTagAndNoteChanges(items: (SNTag | SNNote)[]): void {
+    const notes = items.filter(
+      (item) => item.content_type === ContentType.Note
+    ) as SNNote[];
```
laurentsenta, line 28: With a typeguard you should get the correct type without static casts,

`const isNote = (x: SNItem): x is SNNote => x.content_type === ContentType.Note`

Same for tags, and the call to receiveTagAndNoteChanges in item manager:

```
    // should work and be typesafe.
    const notes: SNote[] = items.filter(isNote );
```

moughxyz: Good call 👍

laurentsenta (approved):

PR #648, opened 2022-03-10
feat: store file mimeType along with name & ext

amanharwara:
## Description

[Internal link](https://app.asana.com/0/1199942776723820/1201931367397968/f)

## Checklist

- [x] This PR has NO tests


moughxyz (approved):

packages/filepicker/src/types.ts
```diff
@@ -7,4 +7,5 @@ export type OnChunkCallback = (
 export type FileSelectionResponse = {
```
moughxyz, line 7: This could probably be replaced with FileMetadata type.

moughxyz: But actually that would require crossing dependencies. So nvm.

